PR #1, opened 2024-11-19
Enable whole testsuite

mcocdawc:
* Enable automatic installation of dependencies, i.e. `libdmet`.

* Enable nearly the whole test suite for GitHub Actions; i.e. I removed most conditional executions based on GITHUB_ACTIONS. Replaced the check for the GITHUB_ACTIONS environment variable with a check for QUEMB_SKIP_EXPENSIVE_TESTS, after all sometimes we also want to skip expensive tests in another environment.

* Moved both packages into subdirectories of a src/ directory.
This way we are forced to actually install for the test-suite and don't miss it anymore, if the automatic build-process gets broken. (As it was the case because of double python_requires arguments)

The GitHub actions are prepared to enable ruff linting and code analysis as part of the test suite:

```yaml
    - name: Check formatting
      run: |
        ruff format --diff || true    # for the moment we want to always report success


    - name: Static analyis
      run: |
        ruff check . || true    # for the moment we want to always report success
```

As soon as the `|| true` is removed we will have linting enforced and conditionally execute the test-suite only if static analysis actually succeeded.

similar to https://github.com/oimeitei/quemb/pull/53


docs/requirements.txt
mscho527, line 1: iirc, readthedocs.io (at least the way it is configured now) pulls only from this file. since the website isn't pulling from this repo for now, it should be fine, but we should keep this in mind.

mscho527 (approved):
Looks good to me!

docs/requirements.txt
mcocdawc: ok, then I would still rather use the `setup.py` as reference and not redundantly define dependencies. (if possible)

.github/workflows/quemb_unittest.yml
```diff
@@ -27,33 +27,46 @@ jobs:
...
+        pip install .
         echo ${{ github.workspace }} > $(python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")/quemb.pth
 
+
+    - name: Check formatting
+      run: |
+        ruff format --diff || true    # for the moment we want to report always report success
+
+
+    - name: Static analyis
```
mscho527, line 58: tiny tiny typo

mscho527: otherwise, the changes look good to me! @mcocdawc

PR #17, opened 2024-11-21
Add more checks

mcocdawc:
- added support for mypy in the CI pipeline

- added support for pylint in the CI pipeline; this is necessary because ruff cannot detect unresolved imports (yet)


mscho527 (approved):
cool nice cool

PR #18, opened 2024-11-21
fixed boolean expressions + bare excepts

mcocdawc:
- Fixed a couple of boolean expressions to not compare with `True/False`, i.e. use
```python
if not A:
# instead of
if A == False:
```

- Fixed some bare excepts


mscho527 (approved):
nice cool cool nice nice

Issue #19, opened 2024-11-21
Scratch dir refactoring


mcocdawc:
The management of the scratch dir is a huge mess at the moment and needs to be streamlined.
In particular there was very strange exception handling when removing the scratch dir in the end.

mcocdawc:
In particualr there is hardcoded
```python
molbe/be_parallel.py:174:        mch.fcisolver.scratchDirectory = "/scratch/oimeitei/" + dname
```

mcocdawc:
#42 is a special case

PR #28, opened 2024-11-22
WIP: Added ScratchManager + start to have some general package for both molbe and kbe

mcocdawc:
Adds a ScratchManager to work on fixing #19 

- [ ] introduce quemb namespace
- [ ] use tempfile mechanism instead of PID


mcocdawc:
This PR was broken up into pieces and merged in the last few hours.

PR #29, opened 2024-11-26
Introduce quemb namespace

mcocdawc:
Introduce a leading `quemb` namespace, i.e. we have now `quemb.kbe` and `quemb.molbe`.

The main reason is to pull out shared functionality into a `quemb.shared` subpackage.


mscho527 (approved):
nice nice nice coooool

PR #30, opened 2024-11-26
Fix for mypy

### LINKED ISSUE #15: make class-level import of functions an explicit method definition
At the moment we have some methods that are defined by importing functions directly into a class name space.
```python
class Example:
    from module import f
```

this is probably better done by wrapping it explicitly into a method.

Something along this line
```python
class Example:
    def f(self, x):
        from module import f as ext_f
        return ext_f(self, x)
```

although in reality it will be a bit more complicated, as we have to ensure that the docstring is also passed on.

mcocdawc:
## don't define methods by importing into class definition body, fixes #15 

This is achieved via different means. 
1. The cleanest (and best) is when the external function was fully transformed into a free function, i.e. not depending on self. Then the method calls the function and passes in the arguments via `self`.
The current implementation of `get_be_error_jacobian` is a prime example.
```python3
from quemb.shared.external.optqn import (
    get_be_error_jacobian as _ext_get_be_error_jacobian,
)

# ...
class # ...

    @copy_docstring(_ext_get_be_error_jacobian)
    def get_be_error_jacobian(self, jac_solver="HF"):
        return _ext_get_be_error_jacobian(self.Nfrag, self.Fobjs, jac_solver)
```

2. Another clean solution: If the "fake-free" function was only used and called in one class, it was just moved to the class definition body.
`molbe.mbe.BE.localize` is an example. (it was previously in `molbe.lo` as a "free" function with a `self` argument.)

3. If the "fake-free" function would require too much boilerplate code, then I created a mixin-class that the `molbe.mbe.BE` and `kbe.pbe.BE` class inherits from. This is ugly, but it makes the previously ugly construct a bit less ugly.

## Fixed `molbe_octane_get_rdms_test.py`

This was fixed by changing to serial execution.
Parallel execution currently fails because of `sratch_dir` mess #19.

## Added helpers for programatically work with docstrings

We now have a `@add_docstring` and a `@copy_docstring` to generate doc-strings for functions from strings, or to copy them from other functions. This is useful for wrapping, if we don't want to copy everything, such as `__name__`, which would be done by `functools.wraps`.

## Fix underscore variable

Moving some functions into the class definition body made a known problem suddenly relevant.
Some variables used double underscore prefixes, to denote "per-fragment-ness"  of a variable. This clashed with the python naming conventions of name-mangling and "private" variables. Fixed some pertinent problems, i.e. renamed `__rdm1` -> `rdm1__`. In the long run we want better names anyway.

## Add the trailing whitespace check also to server-side CI test pipeline

Previously it was only in the commit hook


mcocdawc:
has to be removed in the end
https://github.com/troyvvgroup/quemb/pull/30/commits/fc3e5c0e97c6e318f1d548e72c611f503dd3e7d5#diff-fad60fb90f72865c4f86eb3422bf78dc0bd68bd95cc7528a3e8ba2ff48ce824dR238

src/quemb/kbe/pbe.py
```diff
@@ -12,12 +12,18 @@
...
+from quemb.molbe._opt import BEOPT
 from quemb.shared import be_var
+from quemb.shared.external.optqn import (
+    get_be_error_jacobian as _ext_get_be_error_jacobian,
+)
+from quemb.shared.helper import copy_docstring
 
 
-class BE:
+class BE(Mixin_k_Localize):
```
mscho527, line 26: On a conceptual level, I am a bit confused why `pbe.BE` class would inherit `Mixin_k_Localize`. I know that you are trying to avoid imports that become class methods, but is this conceptually clearer than maybe creating a localized orbital class that `pbe.BE` object initializes within its constructor and keeps as its attribute?

src/quemb/molbe/.DS_Store
mscho527, line 1: lol I guess `.gitignore` is not doing its job...?

mscho527 (commented):
Aside from the comments I had before, I think the PR generally looks good to me. Maybe a bit nitpicky; I wonder what the new class names like `MixinLocalize` mean? I cannot easily see where the Mixin part comes from.

src/quemb/kbe/pbe.py
mcocdawc: (The following should also answer your [question](https://github.com/troyvvgroup/quemb/pull/30#pullrequestreview-2471580161) about the `Mixin` name )

The full point of a "Mixin" class is to provide method(s) to another class, usually via inheritance, while the "Mixin" class itself cannot be instantiated. 

See a better description [here](https://realpython.com/inheritance-composition-python/#mixing-features-with-mixin-classes), [here](https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-is-it-useful), or [here](https://medium.com/@yanxingyang/usage-of-mixin-class-in-python-932b940db80#id_token=eyJhbGciOiJSUzI1NiIsImtpZCI6IjJjOGEyMGFmN2ZjOThmOTdmNDRiMTQyYjRkNWQwODg0ZWIwOTM3YzQiLCJ0eXAiOiJKV1QifQ.eyJpc3MiOiJodHRwczovL2FjY291bnRzLmdvb2dsZS5jb20iLCJhenAiOiIyMTYyOTYwMzU4MzQtazFrNnFlMDYwczJ0cDJhMmphbTRsamRjbXMwMHN0dGcuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJhdWQiOiIyMTYyOTYwMzU4MzQtazFrNnFlMDYwczJ0cDJhMmphbTRsamRjbXMwMHN0dGcuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJzdWIiOiIxMTU3MzY2NDMzMTk3NDg2ODM4NzIiLCJlbWFpbCI6Im9za2FyLndlc2VyQGdtYWlsLmNvbSIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJuYmYiOjE3MzMxNzkzODUsIm5hbWUiOiJPc2thciBXZXNlciIsInBpY3R1cmUiOiJodHRwczovL2xoMy5nb29nbGV1c2VyY29udGVudC5jb20vYS9BQ2c4b2NMV0RzbVhCTV9EalluSW1nVVlWWnNrOXczUFJxQVZ3empJSWVyNXNqRnFHZUg1OVE9czk2LWMiLCJnaXZlbl9uYW1lIjoiT3NrYXIiLCJmYW1pbHlfbmFtZSI6Ildlc2VyIiwiaWF0IjoxNzMzMTc5Njg1LCJleHAiOjE3MzMxODMyODUsImp0aSI6IjBlNjUzYWViY2MzZmNmNDA0ZTE3N2EyZGNhZGE4MjBlZDlkNTJmZTcifQ.fNQbcNL51LhJZeROQgeBe4cUg6FjjU3fVBOOe58I-qN7NioXmSuBcBoceM_F0R31SOiaBgPcLlsP7xuMfoLg0wVDhkutuE5NemaHq37T-98IUmhN__X6XKl3wbdw_XOuZhuCl_BJlXulb150vwQ_i8gia0XnqcLVuvhJbMqSVlkCcufV_wvRQSOXDRkNP-p6Y2xJ1M-q0OcLocLbBvlg6KhyNY7oWrFMvKlk4okM2EdnFOZBRN2_f2E3kiXop1ZObp-Q7RZGKZ9n_qb7-n4u1_pr5mX0tnm95nwutILQHus6-WSTd0pgVPWjQwWyP5WnoAuanB_OkG4-6C5EsUmZFA).


Having a "free" function, that has the argument `self` and actually uses attributes from `self`, is basically the same as having a Mixin class with exactly one method. 
You cannot use this "free" function as a free function on its own.
Modelling this explicitly as a Mixin, makes it a bit more streamlined with general practice. See for example how `pandas` does it [here](https://github.com/pandas-dev/pandas/blob/main/pandas/io/formats/info.py#L841)

> but is this conceptually clearer than maybe creating a localized orbital class that pbe.BE object initializes within its constructor and keeps as its attribute?

I fully agree with your question in general, but it assumes that we can actually meaningfully instantiate a `Localize` class. Not that this is impossible, but at the moment we just have this fake free function and would have to properly refactor. For example we don't want components to show up in both the `BE` and such a `Localize` class.

In the **particular** case of the localize methods  I can actually refactor it away from a Mixin and move it into the `BE` class definition body. It is only imported and used in the `BE` class.

src/quemb/molbe/.DS_Store
mcocdawc: deleted

src/quemb/kbe/pbe.py
mscho527: Oh, I see. Mixins sound similar to abstract classes in C++. Your choice makes much more sense to me now, and I agree with your reasoning. I think it is early to decide whether it makes sense to refactor out localization methods or not. Assuming we revisit this question at a later point (perhaps create an issue for now?), I approve this PR.

mscho527 (approved):
cool cool approve! let's wait for cool @lweisburn to reply 'cool cool'

lweisburn:
> cool cool approve! let's wait for cool @lweisburn to reply 'cool cool'

cool cool

mcocdawc:
> > cool cool approve! let's wait for cool @lweisburn to reply 'cool cool'
> 
> cool cool

nice nice :-D

PR #40, opened 2024-11-27
Add a bit numpy typing

mcocdawc:
- Add a bit of type safety for numpy


mscho527 (approved):
Looks good to me -- I did not know what the `covariant=True` argument was doing at first; maybe a comment with a reference ([like this one](https://stackoverflow.com/questions/61568462/what-does-typevara-b-covariant-true-mean)) could be nice?

PR #47, opened 2024-12-08
Feat: cubegen wrapper for molecular BE

mscho527:
Adds a simple wrapper to write cube files containing embedding orbitals for molecular BE objects


src/quemb/shared/helper.py
```diff
@@ -66,3 +71,34 @@ def ncore_(z: int) -> int:
...
+    be_object : object
+        BE object containing the fragments, each of which contains embedding orbitals.
+    cube_file_path : PathLike
+        Directory to write the cube files to.
+    fragment_idx : Optional[list]
+        Index of the fragments to write the cube files for. If None, all fragments are written.
+    kwargs: Any
+        Keyword arguments passed to cubegen.orbital.
+    """
+    assert type(be_object) is molbe.BE, NotImplementedError(
```
mcocdawc, line 92: # isinstance
I would use `isinstance`, that also works for subclasses.

# assert
In addition, there is one more important thing about `assert` that I did not have time to mention.
`assert` can be stripped off when invoking optimized builds with the `-O` option.
This means:
- It is great to use `assert` for things that we assume to be true in the code. One can even put semi-compute-intensive tests into `assert` to have a slower debug build and a fast optimized release. (The difference is of course more striking for compiled languages.
- It is dangerous to use assert for user validation, if you want it to always be executed. 
- assert will raise an `AssertionError` and accepts a string as second argument. I would not change this behavoiur.

All in all, I would rewrite as:

```python3
if not isinstance(be_object, molbe.BE):
    raise NotImplementedError(...)
```

src/quemb/shared/helper.py
```diff
@@ -66,3 +71,34 @@ def ncore_(z: int) -> int:
...
+        "Support for Periodic BE not implemented yet."
+    )
+    if not fragment_idx:
+        fragment_idx = range(be_object.Nfrag)
+    for idx in fragment_idx:
+        for emb_orb_idx in range(be_object.Fobjs[idx].TA.shape[1]):
+            orbital(
+                be_object.mol,
+                os.path.join(cube_file_path, f"frag_{idx}_orb_{emb_orb_idx}.cube"),
```
mcocdawc, line 101: If cube_file_path is a `pathlib.Path` one can simply write
```python3
cube_file_path / f"frag_{idx}_orb_{emb_orb_idx}.cube"
```
But I understand if you want to keep a more tolerant interface.

tests/molbe_octane_test.py
```diff
@@ -17,7 +20,58 @@
...
+    mybe = BE(mf, fobj)
+
+    # Perform BE density matching.
+    # Uses 20 procs, each fragment calculation assigned OMP_NUM_THREADS to 4
+    # effectively running 5 fragment calculations in parallel
+    mybe.optimize(solver="CCSD", nproc=20, ompnum=4)
+
+    # Compute error
+    be_ecorr = mybe.ebe_tot - mybe.ebe_hf
+    err_ = (ccsd_ecorr - be_ecorr) * 100.0 / ccsd_ecorr
```
mcocdawc, line 44: I know the test should something else, but why not add an assert here for the energy.
It's a cheap thing.

src/quemb/shared/helper.py
```diff
@@ -1,7 +1,12 @@
+import os
 import sys
-from typing import Any, Callable, Optional, TypeVar
+from pyscf.tools.cubegen import orbital
+from quemb import molbe
+from typing import Any, Callable, Optional, TypeVar, Union
+from typing_extensions import TypeAlias
 
 Function = TypeVar("Function", bound=Callable)
+PathLike: TypeAlias = Union[str, os.PathLike]
```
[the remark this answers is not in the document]
mscho527: addressed this -- thanks!

src/quemb/shared/helper.py
mscho527: addressed this -- thanks!

tests/molbe_octane_test.py
mscho527: This one is the existing octane test that we have not activated yet because of the technical issues. It shows as 'changed' in diff because I factored out the repeated `mol, mf` part to a separate function. We should maybe address this when we finally activate this test.

src/quemb/shared/helper.py
mscho527: I may have missed something, but `PathLike` could either be `pathlib.Path` or `str`. The tolerant interface is necessary for `str`, right?

mcocdawc: Yes, if you want to also accept strings you have to either use `os.path.join` or narrow the type down with an explicit cast to `Path`.

Note how the constructor [here](https://github.com/troyvvgroup/quemb/blob/main/src/quemb/shared/manage_scratch.py#L47)
accepts `PathLike`, but then quickly casts to `Path` and internally it is only `Path` as seen [here](https://github.com/troyvvgroup/quemb/blob/main/src/quemb/shared/manage_scratch.py#L44).
This then allows nicer working with pure `Path` as shown [here](https://github.com/troyvvgroup/quemb/blob/main/src/quemb/shared/manage_scratch.py#L108)

My comment is really just a comment; if you want to keep it internally as `PathLike`, go for it.

mcocdawc:
one last thing because it is not yet part of the automatic CI.
could you run ruff `check --fix on` your file?
It should sort the imports.

mcocdawc (approved):
cool cool nice nice

Issue #44, opened 2024-12-06
Remaining errors `ruff check`


mcocdawc:
The following non-trivial errors are reported by `ruff check`:

Do you have any preferences for which file you want to work on, or should we just distribute it evenly?

```
src/quemb/kbe/chain.py:105:27: F821 Undefined name `nsites`

src/quemb/kbe/lo.py:312:37: F821 Undefined name `uciao`

src/quemb/kbe/pbe.py:498:9: F841 Local variable `EH1` is assigned to but never used
src/quemb/kbe/pbe.py:499:9: F841 Local variable `ECOUL` is assigned to but never used
src/quemb/kbe/pbe.py:500:9: F841 Local variable `EF` is assigned to but never used
src/quemb/kbe/pbe.py:672:17: F841 Local variable `eh1` is assigned to but never used
src/quemb/kbe/pbe.py:672:22: F841 Local variable `ecoul` is assigned to but never used
src/quemb/kbe/pbe.py:672:29: F841 Local variable `ef` is assigned to but never used
src/quemb/kbe/pbe.py:899:5: PLC0415 `import` should be at the top-level of a file
src/quemb/kbe/pbe.py:911:5: PLC0415 `import` should be at the top-level of a file

src/quemb/molbe/lchain.py:27:13: F841 Local variable `nk1` is assigned to but never used
src/quemb/molbe/lchain.py:31:9: F841 Local variable `Ns` is assigned to but never used

src/quemb/molbe/mbe.py:558:23: F841 Local variable `rdm1_lo` is assigned to but never used
src/quemb/molbe/mbe.py:558:32: F841 Local variable `rdm2_lo` is assigned to but never used
src/quemb/molbe/mbe.py:894:13: F841 Local variable `dm_init` is assigned to but never used
src/quemb/molbe/mbe.py:915:17: F841 Local variable `eh1` is assigned to but never used
src/quemb/molbe/mbe.py:915:22: F841 Local variable `ecoul` is assigned to but never used
src/quemb/molbe/mbe.py:915:29: F841 Local variable `ef` is assigned to but never used

src/quemb/molbe/pfrag.py:212:9: E722 Do not use bare `except`

src/quemb/molbe/solver.py:102:5: F841 Local variable `E` is assigned to but never used
src/quemb/molbe/solver.py:112:13: F841 Local variable `heff_` is assigned to but never used
src/quemb/molbe/solver.py:190:13: F841 Local variable `nao` is assigned to but never used
[...]

PR #48, opened 2024-12-09
Fix: Address remaining ruff errors (cphf_utils & optqn)

mscho527:
```
src/quemb/shared/external/cphf_utils.py:34:5: F841 Local variable `nv` is assigned to but never used
src/quemb/shared/external/cphf_utils.py:43:5: F841 Local variable `nv` is assigned to but never used
src/quemb/shared/external/cphf_utils.py:384:5: F841 Local variable `nov` is assigned to but never used
src/quemb/shared/external/cphf_utils.py:403:5: F841 Local variable `nov` is assigned to but never used
```
All flagged variables `nv & nov` are the number of virtuals and the number of (occupied + virtuals) that are not used in the code.

```
src/quemb/shared/external/optqn.py:90:26: F821 Undefined name `ared`
src/quemb/shared/external/optqn.py:210:13: F841 Local variable `us_tmp` is assigned to but never used
src/quemb/shared/external/optqn.py:219:13: F841 Local variable `alp` is assigned to but never used
src/quemb/shared/external/optqn.py:404:10: F841 Local variable `Jcs` is assigned to but never used
```
`ared` is a variable later defined in the loop, but not accessed in line 90 due to short-circuiting. For code style compliance, `ared` is now pre-defined with a value.
`us_tmp, alp, Jcs` are safely removed unused variables.

part of #44

### DIFF
--- a/src/quemb/shared/external/optqn.py
+++ b/src/quemb/shared/external/optqn.py
@@ -87,6 +87,7 @@ def trustRegion(func, xold, fold, Binv, c=0.5):
     dx_sd = -B.T @ fold  # Steepest Descent step
     t = numpy.linalg.norm(dx_sd) ** 2 / numpy.linalg.norm(B @ dx_sd) ** 2
     prevdx = None
+    ared = 0.0  # Reduction in the objective function; Initialize value to 0
     while ratio < rho or ared < 0.0:
         # Trust Region subproblem
         # minimize (1/2) ||F_k + B_k d||^2 w.r.t. d, s.t. d w/i trust radius
@@ -207,7 +208,6 @@ class FrankQN:
                 dx_i @ self.Binv @ df_i
             )
             self.Binv += tmp__
-            us_tmp = self.Binv @ self.fnew
 
         if trust_region:
             self.xnew, self.fnew = trustRegion(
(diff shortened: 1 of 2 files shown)


src/quemb/shared/external/optqn.py
```diff
@@ -388,7 +388,6 @@ def get_atbe_Jblock_frag(fobj, res_func):
 
 def get_be_error_jacobian_selffrag(self, jac_solver="HF"):
     Jes = [None] * self.Nfrag
-    Jcs = [None] * self.Nfrag
     xes = [None] * self.Nfrag
     xcs = [None] * self.Nfrag
     ys = [None] * self.Nfrag
```
mcocdawc, line 393: if you know the correct datatype (i.e. float, int, or complex) then it would be great to also replace the `[None] * self.Nfrag` with `np.empty` calls.

If you are not sure about it, then we just leave it as is.

mscho527: Hmm I do know the datatypes, but do not know what their sizes are off the top of my head :(

mcocdawc: Shouldn't the size just be `self.Nfrag`?
i.e. 
```python3
np.empty(self.Nfrag, dtype=??)
```

mscho527: The elements of the `Jes`, for example, would be numpy arrays, which I do not know the size of. Do you mean I should do `np.empty(self.Nfrag, dtype=np.ndarray`)?

mcocdawc: Summarizing the discussion with @mscho527 : 

with ragged arrays it becomes too complicated now. let's postpone it for later

mcocdawc (approved):

PR #49, opened 2024-12-09
Manual fix ruff

mcocdawc:
Fixed remaining errors in `src/quemb/shared/external/cpmp2_utils.py` and `src/quemb/molbe/solver.py` 

- removed unused variables
- made it explicit that `update_heff` modifies self, i.e. it returns nothing now.
- removed `ecore` from everywhere. It did not do anything anyway. If it should be reintroduced it is much easier, if automatic code analysis (unused arguments in particular) works.

part of #44


mscho527:
> * removed `ecore` from everywhere. It did not do anything anyway. If it should be reintroduced it is much easier, if automatic code analysis (unused arguments in particular) works.

It looks like functionality related to frozen core was stripped down here; we should perhaps discuss how to proceed with this.

mcocdawc:
> > * removed `ecore` from everywhere. It did not do anything anyway. If it should be reintroduced it is much easier, if automatic code analysis (unused arguments in particular) works.
> 
> It looks like functionality related to frozen core was stripped down here; we should perhaps discuss how to proceed with this.

the functionality does not exist. There is an `ecore` argument to the outside, but that is never used. I would rather explicitly remove it and then cleanly add it, than have a non-working argument.

src/quemb/molbe/solver.py
```diff
@@ -212,7 +205,7 @@ def be_func(
             nmo = fobj._mf.mo_coeff.shape[1]
             nelec = (fobj.nsocc, fobj.nsocc)
             cas = mcscf.CASCI(fobj._mf, nmo, nelec)
-            h1, ecore = cas.get_h1eff(mo_coeff=fobj._mf.mo_coeff)
+            h1, _ = cas.get_h1eff(mo_coeff=fobj._mf.mo_coeff)
```
mscho527, line 208:
```suggestion
            h1, _ = cas.get_h1eff(mo_coeff=fobj._mf.mo_coeff)  # h1, ecore
```

src/quemb/molbe/solver.py
```diff
@@ -400,8 +390,6 @@ def be_func_u(
 
     # Loop over each fragment and solve using the specified solver
     for fobj_a, fobj_b in Fobjs:
-        heff_ = None  # No outside chemical potential implemented for unrestricted yet
```
mscho527, line 403: we probably want to create an issue @lweisburn

mscho527 (changes requested):
mostly cool, i think a few comments would be nice!

mscho527 (approved):

PR #50, opened 2024-12-09
Fix docs

### LINKED ISSUE #5: Escalate warnings to errors when building the documentation
Eventually we want to have
`make html SPHINXOPTS="-W --keep-going -n"`
when building the documentation in the test-suite.
This first requires to fix all the existing warnings. :-(

mcocdawc:
- added improvements to the documentation, such as links to external libraries
- Fixed the broken links in the documentation
- actually fail the build when links are broken (this is shown in [this build](https://github.com/troyvvgroup/quemb/actions/runs/12246349558/job/34162068173?pr=50) where I consciously added a broken link)
- changed an `assert` into if-statement in `src/quemb/molbe/misc.py`, since it should also be there in production code

Fixes #5


docs/source/solvers.rst
```diff
@@ -5,15 +5,6 @@ Solver Routines
...
 
-Molecular orbital localization
-------------------------------
-
-.. autofunction:: quemb.molbe.lo.localize
-
-Crystalline orbital localization
---------------------------------
-
-.. autofunction:: quemb.kbe.lo.localize
```
mscho527, line 16: Maybe I'm missing something -- is there a reason why we are removing the autodocs here (and 67-74 below)?

src/quemb/molbe/be_parallel.py
```diff
@@ -609,7 +609,7 @@ def be_func_parallel_u(
         Potentials (local & global) that are added to the 1-electron
         Hamiltonian component.  The last element in the list is the chemical potential.
         Should always be 0, as this is still a one-shot only implementation
-    Fobjs : list of tuples of MolBE.fragpart
+    Fobjs : list of tuple of quemb.molbe.fragment.fragpart
```
mscho527, line 612: actually, i don't think Fobjs can be list of tuples..? @lweisburn any thoughts?

mscho527 (changes requested):

docs/source/solvers.rst
mcocdawc: these functions do not exist anymore they are now part of the `BE` class and moved there as methods.
(They are documented via `autoclass`)

This was done in #30

mscho527: ah i see! cool!

src/quemb/molbe/be_parallel.py
mscho527: my bad -- it should be list of tuple

mscho527 (approved):
```
                                                      ______
_________________________________________________________  /
_  ___/  __ \  __ \  __ \  __ \  __ \  __ \  __ \  __ \_  / 
/ /__ / /_/ / /_/ / /_/ / /_/ / /_/ / /_/ / /_/ / /_/ /  /  
\___/ \____/\____/\____/\____/\____/\____/\____/\____//_/   
                                                            
```

PR #56, opened 2024-12-10
Fixes issues identified by `ruff` and `flake8` in `pbe.py`

ShaunWeatherly:
Addresses warnings of type `F841` and `PLC0415` in module `pbe.py`. Also changes an ambiguous variable name `I` to the more descript`fidx`.

Full breakdown:
src/quemb/kbe/pbe.py:498:9: F841 Local variable `EH1` is assigned to but never used
src/quemb/kbe/pbe.py:499:9: F841 Local variable `ECOUL` is assigned to but never used
src/quemb/kbe/pbe.py:500:9: F841 Local variable `EF` is assigned to but never used
src/quemb/kbe/pbe.py:672:17: F841 Local variable `eh1` is assigned to but never used
src/quemb/kbe/pbe.py:672:22: F841 Local variable `ecoul` is assigned to but never used
src/quemb/kbe/pbe.py:672:29: F841 Local variable `ef` is assigned to but never used
src/quemb/kbe/pbe.py:899:5: PLC0415 `import` should be at the top-level of a file
src/quemb/kbe/pbe.py:911:5: PLC0415 `import` should be at the top-level of a file


src/quemb/kbe/pbe.py
```diff
@@ -665,7 +663,7 @@ def initialize(self, eri_, compute_hf, restart=False):
 
             # energy
             if compute_hf:
-                eh1, ecoul, ef = self.Fobjs[frg].energy_hf(return_e1=True)
+                self.Fobjs[frg].energy_hf(return_e1=True)
```
mcocdawc, line 666: it would be a great idea to refactor `energy_hf` to actually return nothing.
After all it is a method with side-effects on `self` and this would make it explicit.

(If this is not possible because the returned value is actually used somewhere else, then I would set at least `return_e1=False` here)

src/quemb/kbe/pbe.py
```diff
@@ -855,8 +853,8 @@ def initialize_pot(Nfrag, edge_idx):
     pot_ = []
 
     if not len(edge_idx) == 0:
-        for I in range(Nfrag):
```
mcocdawc, line 858: I am fine with this change, but note that we consciously deactivated [ambiguous-variable-name (E741)](https://docs.astral.sh/ruff/rules/ambiguous-variable-name/#ambiguous-variable-name-e741) [here](https://github.com/troyvvgroup/quemb/blob/main/.ruff.toml#L9)

It is just very common to have these letters in mathematical equations, hence in our code.

mscho527:
Resolves #44

mcocdawc (approved):
cool cool nice nice

PR #66, opened 2024-12-13
Add config

mcocdawc:
This PR streamlines the `be_var` configuration to a better name and introduces an optional yaml configuration file `~/.quembrc`.


ShaunWeatherly (approved):
This looks good, I much prefer your `settings` naming scheme over the old one. This also makes it super straightforward to add new `settings` attributes in the future.

PR #70, opened 2024-12-17
Automatic generation of API documentation

### LINKED ISSUE #65: make the documentation follow the namespace structure
The documentation is organized completely orthogonal to the namespace structure.
At least the API description should just follow the namespace ordering.
This would allow us to also use more automatic tooling for the generation of `rst` files, such as `sphinx-autogen`.

In addition, we could give some tutorial examples.

mcocdawc:
- Create API reference automatically for all public members of `quemb`
     
    - Before this change one had to manually create `rst` files and manage them. Now they are recursively created automatically.
    - The layout of the documentation matches the layout of the namespace

- This change automatically ensures that all (public) docstrings are actually parsed by Sphinx and it is ensured that they are properly written. Had to fix some docstrings

- Type-hinting is now picked up by sphinx.
```python
def double(n: int):
    """
    Parameters
    -----------------
    n: int
        fancy natural number
    """
    return 2 * n
```
is not necessary, one can omit the redundant `int` and write:
```python
def double(n: int):
    """
    Parameters
    -----------------
    n:
        fancy natural number
    """
    return 2 * n
```

Fixes #65


src/quemb/kbe/lo_k.py
```diff
@@ -140,12 +154,20 @@ def get_iao_k(Co, S12, S1, S2=None, ortho=True):
...
-        Ciao: output of :func:`get_iao`
-        S: ao ovlp matrix
-        S12: valence orbitals projected into ao basis
-    Return:
-        Cpao (orthogonalized)
+
+    Parameters
+    ----------
+    Ciao:
+        output of :code:`get_iao`
```
mcocdawc, line 161: @lweisburn to which `get_iao` should this refer?

src/quemb/kbe/lo_k.py
```diff
@@ -163,13 +185,22 @@ def get_pao_k(Ciao, S, S12):
...
-        S: ao ovlp matrix
-        mol: mol object
-        valence basis: basis used for valence orbitals
-    Return:
-        Cpao (symmetrically orthogonalized)
+
+    Parameters
+    ----------
+    Ciao :
+        output of :code:`get_iao_native`
```
mcocdawc, line 192: @lweisburn to which `get_iao_native` should this refer?

src/quemb/kbe/lo_k.py
lweisburn: I think this should be `get_iao_k` in `kbe/lo_k`, which gets iao's across k-points, instead of the `get_iao` in `molbe`

src/quemb/kbe/lo_k.py
lweisburn: `get_iao_native` is an option in frankenstein, but not quemb. The `get_iao_native` I _think_ should be the same as `get_iao` in `molbe/lo`, but I am not sure and need to actually test that

ShaunWeatherly (approved):
This looks good, it'll force us to write clean and complete docstrings. I do think the in-line text formatting (ie :python: ...) is a bit ugly, but I get the utility and I can't think of a better way to do it...

mcocdawc:
> This looks good, it'll force us to write clean and complete docstrings. I do think the in-line text formatting (ie :python: ...) is a bit ugly, but I get the utility and I can't think of a better way to do it...

yeah I don't like it either, but that is the sphinx syntax :-(

src/quemb/kbe/lo_k.py
mcocdawc: makes sense and changed, thank you :)

src/quemb/kbe/lo_k.py
mcocdawc: In this case I would just leave it as 
```
:code:`get_iao_native`
```
and merge. This way it is rendered as code, but does not link to any function in the code.
That is not an ideal situation, but it was the way the code was before and I think it makes no sense to postpone this PR for it.

Are you ok with that?

lweisburn: I am okay with this for now! We will have to go through the IAO code more thoroughly later on

Issue #35, opened 2024-11-27
`kbe.pbe.BE.compute_energy_full` is called, but never defined


mcocdawc:
`kbe.pbe.BE.compute_energy_full` is called, but never defined

the call is in `kbe.pbe.BEoneshot` and looks like this
```python
        if not calc_frag_energy:
            self.compute_energy_full(approx_cumulant=True, return_rdm=False)
```

(The problem does not exist in `molbe`.)

PR #71, opened 2024-12-18
WIP: Fix scratch dir

mcocdawc:


src/quemb/kbe/pbe.py
```diff
@@ -759,15 +760,7 @@ def oneshot(
 
             self.ebe_tot = rets[0]
 
-        if not calc_frag_energy:
```
mcocdawc, line 762: This call to `compute_energy_full` here is removed, because it is not existing

#35 is related

src/quemb/molbe/be_parallel.py
```diff
@@ -460,12 +462,11 @@ def be_func_parallel(
...
         the error norm, error vector, and the computed energy.
     """
-    nfrag = len(Fobjs)
     # Create directories for fragments if required
     if writeh1 and solver == "SCI":
-        for nf in range(nfrag):
-            dname = Fobjs[nf].dname
-            os.system("mkdir " + dname)
+        for fobj in Fobjs:
+            # TODO: continue here
```
mcocdawc, line 468: continue work from here

mcocdawc:
did a cleaner attempt in #74

PR #74, opened 2024-12-19
Final scratch dir attempt

### LINKED ISSUE #19: Scratch dir refactoring

mcocdawc:
# Mayor changes

This PR unifies the use of the ScratchDir. It is now guaranteed, that files are cleaned up if the calculation exits with no error. (If the user passes `WorkDir(cleanup_at_end=False)` the files are retained).
A `WorkDir` is passed from the top of the call stack to the bottom. This allows the user to change the WorkDir if necessary. (Fixes #19)

Got rid of some special cased keyword arguments for DMRG and moved them into an explicit `DMRG_solver_kwargs`.
This way we can catch arguments that are supplied and do nothing.
It is probably a good idea to do the same for SHCI specific keywords @lweisburn .
It is probably a good idea to get rid of the dictionary altogether and to replace it with a dataclass.

# Minor changes

If the `__init__` of a class was just populating the attributes, then I replaced the class definition with `attrs.define`.

On the way I did a lot of type hinting. It is probably a good idea to double check my assumptions in review.

Additionally, I did some refactoring, where explicit open-close pairs where replaced with context managers.
The same was true for multiprocessing Pools.

Here and there I encountered constructs that could be simplified a lot
```python3
veffs = []
[veffs.append(result.get()) for result in results]
# becomes
veffs = [result.get() for result in results]
```
or
```python3
self.frozen_core = False if not fobj.frozen_core else True
# becomes
self.frozen_core = fobj.frozen_core
```
or
```python3
rdm_return = False
if relax_density:
    rdm_return = True
# becomes
rdm_return = relax_density
```


src/quemb/molbe/mbe.py
```diff
@@ -19,42 +21,28 @@
...
-        self.hf_etot = hf_etot
-        self.W = W
-        self.lmo_coeff = lmo_coeff
-        self.enuc = enuc
-        self.E_core = E_core
-        self.C_core = C_core
-        self.P_core = P_core
-        self.core_veff = core_veff
-        self.mo_energy = mo_energy
+    # TODO: some of the types are most likely wrong.
```
mcocdawc, line 29: these types have to be checked in a review

mcocdawc:
I added both you @lweisburn and @ShaunWeatherly as reviewers and will not merge gefore there are two approvals, since it is such a long PR.

src/quemb/shared/manage_scratch.py
```diff
@@ -65,7 +105,7 @@ def __exit__(
         value: BaseException | None,
         traceback: TracebackType | None,
     ) -> Literal[False]:
-        if value is None:
+        if self.cleanup_at_end:
```
mcocdawc, line 108: this change might be relevant for you @lweisburn .

If the WorkDir is used as a contextmanager, then the files are deleted upon exiting the with-statement, even if errors occur.

This is different to the behavior, when calling it in the normal way, where the `atexit.register` means that the WorkDir is only cleaned, if the calculation exits cleanly.

Since you can now pass a WorkDir to BE and it is passed all the way down, this means you can have a script and multiple BE calculations that are guaranteed to be cleaned up, even if errors occur if you write:

```python3
with WorkDir() as scratch:
    BE(..., scratch_dir=scratch)
# here scratch is deleted, even if errors happen.
```
while the normal way of writing it:
```python3
BE(...)
```
cleans the files only upon a clean exit.

tests/scratch_manager_test.py
```diff
@@ -27,7 +27,7 @@ def test_keep_upon_error() -> None:
     with raises(ValueError):
         with WorkDir(my_tmp):
             raise ValueError
-    assert my_tmp.exists()
+    assert not my_tmp.exists()
```
mcocdawc, line 30: @lweisburn just to show how the scratch is deleted even if errors occur.

src/quemb/molbe/mbe.py
```diff
@@ -735,15 +712,15 @@ def optimize(
...
+            DMRG_solver_kwargs=DMRG_solver_kwargs,
         )
 
         if method == "QN":
             # Prepare the initial Jacobian matrix
             if only_chem:
                 J0 = [[0.0]]
                 J0 = self.get_be_error_jacobian(jac_solver="HF")
-                J0 = [[J0[-1, -1]]]
+                J0 = [[J0[-1][-1]]]
```
mcocdawc, line 723: BTW: this error was found because of typing.

src/quemb/shared/manage_scratch.py
lweisburn: How cool!! This will be incredibly helpful, thank you!

src/quemb/molbe/mbe.py
mscho527: afaik, `J0[-1][-1]` and `J0[-1, -1]` should both work fine, since `get_be_error_jacobian` returns a `numpy.ndarray`. Type declaration on Line 738 has `list[list[float]]`, whereas what's returned from `get_be_error_jacobian` is defined as `numpy.ndarray` [here](https://github.com/troyvvgroup/quemb/blob/2e934d8be2844e605b9c37aeae0e1ae0c8371585/src/quemb/shared/external/optqn.py#L288).

mscho527: actually, since `J0` is a `numpy.ndarray`, `[[J0[-1][-1]]` would be slightly inefficient since a copy of `J0[-1]` will be created before it is indexed. i think this line should be reverted back to `[-1, -1]` and the type declaration on 738 should be modified instead.

partial credit for @lweisburn for raising question on this

src/quemb/molbe/mbe.py
```diff
@@ -1,60 +1,48 @@
...
+    S: Matrix[float64]
+    C: Matrix[float64]
+    hf_dm: Matrix[float64]
+    hf_etot: float
+    W: Matrix[float64]
+    lmo_coeff: Matrix[float64]
+    enuc: float
+    ek: float
+    E_core: float
+    C_core: float
```
lweisburn, line 43: I think C_core, P_core, and core_veff should all be Matrix[float64], not floats

src/quemb/kbe/misc.py
```diff
@@ -31,40 +34,23 @@ def get_phase1(cell, kpts, kmesh):
...
+    hcore: Matrix[float64]
+    S: Matrix[float64]
+    C: Matrix[float64]
+    hf_dm: Matrix[float64]
+    hf_etot: float
+    W: Matrix[float64]
+    lmo_coeff: Matrix[float64]
+    enuc: float
+    ek: float
+    E_core: float
```
lweisburn, line 50: Same comment here for P_core, C_core, and core_veff: all should be Matrix[float64], not floats

lweisburn:
Looks good to me!

src/quemb/molbe/solver.py
```diff
@@ -17,30 +20,32 @@
...
-    select_cutoff=None,
-    eeval=False,
-    ereturn=False,
-    frag_energy=False,
-    relax_density=False,
-    return_vec=False,
-    use_cumulant=True,
-    scratch_dir=None,
-    **solver_kwargs,
+    pot: list[float] | None,
```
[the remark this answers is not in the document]
mcocdawc: resolved after talking directly with @lweisburn

src/quemb/kbe/pbe.py
```diff
@@ -47,65 +49,67 @@ class BE(Mixin_k_Localize):
...
-        ompnum=4,
-        hci_cutoff=0.001,
-        ci_coeff_cutoff=None,
-        select_cutoff=None,
-        iao_val_core=True,
-        exxdiv="ewald",
-        kpts=None,
-        cderi=None,
-        iao_wannier=False,
+        mf: pbc.scf.hf.SCF,
```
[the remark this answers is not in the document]
mcocdawc: resolved after talking directly with @lweisburn

src/quemb/molbe/mbe.py
```diff
@@ -80,68 +68,68 @@ class BE(MixinLocalize):
...
-        nproc=1,
-        ompnum=4,
-        scratch_dir=None,
-        hci_cutoff=0.001,
-        ci_coeff_cutoff=None,
-        select_cutoff=None,
-        integral_direct_DF=False,
-        auxbasis=None,
-    ):
+        mf: scf.hf.SCF,
```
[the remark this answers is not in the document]
mcocdawc: resolved after talking directly with @lweisburn

lweisburn (approved):
The changes look good to me!

src/quemb/molbe/mbe.py
mcocdawc: resolved after talking to leah

mcocdawc:
talking directly to @ShaunWeatherly , he is also fine with it and I proceed with merging.

Issue #72, opened 2024-12-18
working parallelization for different OS's


lweisburn:
running `be_func_parallel` (i.e. request >1 nproc) uses `multiprocessing.Pool`. This has a different default depending on operating system, but only `fork` works with our code (not `spawn`, the default with iOS, and maybe not the newer standard `forkserver`)

To fix, we need to simply specify explicitly that `multiprocessing` uses `fork`:
`multiprocessing.set_start_method('fork')`

mscho527:
Did #75 fully fix this issue for now? If so, we could maybe close this issue.

mcocdawc:
#75 only made the tests working. The question about multiprocessing and different OSs is not yet entirely solved.

Just my 2cts:

- If we want to support more than Linux we have to move way from fork-based multiprocessing.

- Even if we stick to Linux and fork-based multiprocessing, I think there are some wrong assumptions in the code about how forking works. In this call to [apply_async](https://github.com/troyvvgroup/quemb/blob/main/src/quemb/molbe/be_parallel.py#L484) there is a copy of an argument. This suggests that the author thought that the other arguments are not copied, which is unfortunately not true. The passed arguments are always pickled, sent to the new process and unpickled, i.e. copied. Only module level variables are shared, **in the fork-sense of copy-on-write**. **At the moment all arguments are copied, i.e. if there are n processes, this increases the memory demand for the arguments by a factor of n.** If we use multiprocessing we should pass explicitly shared memory for expensive read-only arguments.

- Thread-based parallelization is more dangerous, but works on Windows and seems to work wit the limited amount of tests. See my [draft PR here](https://github.com/troyvvgroup/quemb/pull/78)

lweisburn:
As @mcocdawc says ^^ !

#75 quickly fixed the code such that oneshot BE and regular BE work on multiple cores, as before the code was moved here. It also fixed and enabled the test suite for these calculations. It doesn't make any fix for non-Linux or explore different parallelization options.

As for the copying: the objects passed through multiprocessing were intentionally not the expensive ones (i.e., passing the ERI file location, not the full ERIs) because of the copies. The hf_veff matrix and 1 electron integrals are the largest I think. I don't think that we need to worry about memory in this step yet, though we could explore other options in the future? To my understanding, this behavior is the case with all the standard python multiprocessing.. ?

mscho527:
[...]

Issue #33, opened 2024-11-27
FCI Approx Cumulant


mscho527:
_(Per discussion w/ @lweisburn)_

As it is currently implemented, it looks like we calculate the exact cumulant for the FCI solver and use the energy expression for approximate cumulant. Investigate if the mismatch causes particular issues, and resolve the mismatch.

mcocdawc:
Perhaps I am misinterpreting stuff here; but the following might be related:

The energy is exactly the same, regardless of the argument to `approx_cumulant` in
```python3
be.compute_energy_full(approx_cumulant=...
```

https://github.com/troyvvgroup/quemb/pull/30/files#diff-cfd62b5afc590cdc497c40dba3ec4f6bb7c2c05fac6e7ba6f4fbf7fc561e27a7R67

mscho527:
It might be related; we realized this while looking at the `Frag.energy` function (which we no longer call in this version of the code), but I haven't looked in detail how the `compute_energy_full` and/or `get_frag_energy` may or may not be affected in the same fashion. Will update soon after scrutinizing!

lweisburn:
Summarizing our @mscho527 conversation this week and some more digging about cumulant energy forms in the code for future reference:

In the CCSD solver (in the normal `get_frag_energy` function), we use `pyscf`'s rdm2 build with the `with_dm1=False` option. This is equivalent to what we call the approximate cumulant (i.e. does _not_ have the \delta P \delta P terms), so we use it with the approximate cumulant energy expression (exactly as written in the periodic BE paper). We do not have the true cumulant version included with `get_frag_energy`, which would require an updated fock matrix for the full system. We could add this, but the approximate cumulant works well here.

The current cumulant option for FCI is also the approximate cumulant, so it should be as-is correct with `get_frag_energy`.  Again, we could easily update this function to be the true cumulant in the future.

in all: we need to update `get_frag_energy` if we want to use the true cumulant instead (which I think we can do, and I could do after the break). 

TODO, either later or after break: go back through the energy and full system rdm construction

lweisburn:
Looking at the `compute_energy_full`: this should be returning (a very slightly different) true and approximate cumulant energy correctly, right?

[...]

PR #75, opened 2024-12-19
Fix parallelization and add parallel tests

lweisburn:
Made small changes to `_opt` to correctly interface with `be_parallel`

TODO: Fix and check examples and tests to make sure we have a functional test suite. After this is done, we can work on making the parallel code work on different operating systems (Issue #72)

### DIFF
--- a/tests/molbe_octane_test.py
+++ b/tests/molbe_octane_test.py
@@ -12,14 +12,6 @@ from pyscf import cc, gto, scf
 from quemb.molbe import BE, fragpart
 from quemb.shared.io import write_cube
 
-# TODO: actually add meaningful tests for energies etc.
-#   At the moment the test fails already for technical reasons.
-
-
-@pytest.mark.skipif(
-    not os.getenv("QUEMB_DO_KNOWN_TO_FAIL_TESTS") == "true",
-    reason="This test is known to fail.",
-)
 def test_octane_molbe() -> None:
     # Prepare octane molecule
     mol, mf = prepare_octane()
--- a/tests/molbe_octane_test.py
+++ b/tests/molbe_octane_test.py
@@ -23,19 +23,23 @@ def test_octane_molbe() -> None:
     print(f"*** CCSD Correlation Energy: {ccsd_ecorr:>14.8f} Ha", flush=True)
 
     # initialize fragments (use frozen core approximation)
-    fobj = fragpart(be_type="be2", mol=mol, frozen_core=True)
+    fobj = fragpart(be_type="be2", mol=mol, frozen_core=False)
     # Initialize BE
     mybe = BE(mf, fobj)
 
     # Perform BE density matching.
-    # Uses 20 procs, each fragment calculation assigned OMP_NUM_THREADS to 4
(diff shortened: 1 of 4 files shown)


lweisburn:
This at least allows for >1 nproc to be requested for standard oneshot and iterative BE calculations. It also modifies a two of the tests to use the parallel version of the code. 

This sets us up for an updated type of parallelization (Issue #72). Note that `be_func_parallel` and other tests still need to be "cleaned up" and keywords standardized, especially with respect to energy and rdm returns (somewhat in Issue #33, #35)

tests/molbe_octane_test.py
```diff
@@ -6,20 +6,12 @@
 from typing import Tuple
 
 import numpy as np
-import pytest
 from pyscf import cc, gto, scf
 
 from quemb.molbe import BE, fragpart
 from quemb.shared.io import write_cube
 
-# TODO: actually add meaningful tests for energies etc.
```
mcocdawc, line 15: Since the test runs now, I would add meaningful tests for energies etc.
(Basically what the TODO says :-D)

This can really be as simple as adding
```
assert np.isclose(mybe.ebe_tot, ...)
assert np.isclose(mybe.ebe_hf, ...)
```
where you get the expected value by simply printing the current value.

Basically we assume it is correct now and want to detect deviations from it.

tests/molbe_octane_test.py
mcocdawc, line 1: At the moment we have:
```python
    # Perform BE density matching.
    # Uses 20 procs, each fragment calculation assigned OMP_NUM_THREADS to 4
    # effectively running 5 fragment calculations in parallel
    mybe.optimize(solver="CCSD", nproc=20, ompnum=4)
```

I am surprised that even ran on the GitHub Actions.

We should have a non-trivial parallelized but cheap test that can be verified even on bad hardware.
I.e. something closer to `nproc=4, ompnum=2`

(Assuming that the test does not test behaviour that only appears for `nproc=>20`)

tests/molbe_octane_test.py
```diff
@@ -6,20 +6,12 @@
...
-#   At the moment the test fails already for technical reasons.
```
mcocdawc, line 16: Since the test runs now, I would add meaningful tests for energies etc.
(Basically what the TODO says :-D)

This can really be as simple as adding
```
assert np.isclose(mybe.ebe_tot, ...)
assert np.isclose(mybe.ebe_hf, ...)
```
where you get the expected value by simply printing the current value.

Basically we assume it is correct now and want to detect deviations from it.

tests/molbe_octane_test.py
lweisburn: Sure, I'll make that change!

tests/molbe_octane_test.py
lweisburn: Great! I added these exactly

mcocdawc (approved):
ready to go

```text
        |
       / \
      / _ \
     |.o '.|
     |'._.'|
     |     |
   ,'|  |  |`.
  /  |  |  |  \
  |,-'--|--'-.| 
```

Issue #60, opened 2024-12-11
Test IAO


mscho527:
Test IAOs.

Once we know that it works, add arg to accept iao_basis in `molbe.misc.py:be2puffin`

mscho527:
`lo.py:get_pao` makes no sense at this point (touches numpy namespace). Fix this

mscho527:
`lo_k.py:get_pao` as well

mscho527:
As noted in the review comment in #76, reordering by atom might be unnecessary. Reminder to verify this when we test IAOs

lweisburn:
Much of this is fixed with #102 . A similar review with the periodic code is coming, as well as an iao_basis argument in `molbe.misc.py:be2puffin`

cariluo:
Could you please clarify what is left to be done on this issue? (I have done some preliminary force embedding calculations with extended basis sets and IAOs and may do more in the future.)

mscho527:
I think leah's last comment is still where we are at with iaos!

PR #76, opened 2024-12-23
More type annotations + simpler boolean expressions

mcocdawc:
Simplifications in the code

- A couple more type annotations

- simpler boolean expressions

- removed keyword arguments if they are always true in the code


src/quemb/kbe/lo.py
```diff
@@ -118,19 +118,15 @@ def localize(
...
-                else:
-                    Ciao_ = ciao_.copy()
+                nk, nao, nlo = ciao_.shape
+                Ciao_ = numpy.zeros((nk, nao, nlo), dtype=numpy.complex128)
+                for k in range(self.nkpt):
+                    aoind_by_atom = get_aoind_by_atom(self.cell)
+                    ctmp, iaoind_by_atom = reorder_by_atom_(
+                        ciao_[k], aoind_by_atom, self.S[k]
+                    )
+                    Ciao_[k] = ctmp
```
mscho527, line 129: As the comment suggests, reordering might be unnecessary. Verify this when dealing with #60

src/quemb/molbe/misc.py
```diff
@@ -393,12 +392,11 @@ def be2puffin(
                 mf = qmmm.mm_charge(
                     mf1, pts_and_charges[0], pts_and_charges[1], unit="bohr"
                 ).newton()
-                print(
+
+                raise ValueError(
                     "Setting use_df to false and jk to none: have not tested DF and "
                     "QM/MM from point charges at the same time"
                 )
```
mscho527, line 399:
```suggestion
                if not use_df or not jk:
                    raise ValueError(
                        "Setting use_df to false and jk to none: have not tested DF and "
                        "QM/MM from point charges at the same time"
                    )
```

mscho527: When `use_df` is False and `jk` is None, we want QM/MM to run

mscho527 (changes requested):
Looks good otherwise!

src/quemb/kbe/lo.py
mcocdawc: but for the moment I leave it it as is?

src/quemb/molbe/misc.py
mcocdawc: your change is different from what you are saying in words.
```python3
if not use_df and jk is None:
    # run
```
becomes
```python3
if not (not use_df and jk is None):
    # raise error
```
becomes
```python3
if use_df or jk is not None:
    # raise error
```

what do we want?

src/quemb/kbe/lo.py
mscho527: yes yes! Sorry should've been clearer :)

src/quemb/molbe/misc.py
mscho527: The last code block would do! I didn't notice this before, but I suppose this input validity check & exception should happen before actually doing the mf calculations.

mcocdawc: thank you and done

mscho527 (approved):
cool cool coool niice

mscho527 (changes requested):
@mcocdawc leah has a comment that we should probably consider before merging pr; un-approving this for now

mcocdawc:
> @mcocdawc leah has a comment that we should probably consider before merging pr; un-approving this for now

oh yeah, I wanted to ask for her review anyway due to my merge conflicts with her changes.

lweisburn:
I had a comment about `return_rdm`, per a discussion with @mscho527, that I want to rescind! The current logic in `be_parallel` makes sense to me wrt to the rdm returns and relaxed/unrelaxed densities. I think that in the future, if we add something to the code such that we do not need the energy and/or the rdms at each iteration, we will have to make some changes, but this makes sense for now!

src/quemb/molbe/lo.py
```diff
@@ -161,12 +173,10 @@ def get_pao(Ciao, S, S12):
     Cpao = (numpy.eye(n) - Piao) @ nonval  # project out IAOs from non-valence basis
 
     # begin canonical orthogonalization to get rid of redundant orbitals
-    Cpao = cano_orth(Cpao, ovlp=S)
-
-    return Cpao
+    return cano_orth(Cpao, ovlp=S)
```
lweisburn, line 176: I think this could be misleading: get_pao is returning a function of Cpao (which we really want to be calling cpao). Instead, maybe we call Cpao in line 173 Cpao_redundant or something? Cpao should be refering to the canonically orthogonalized version

src/quemb/molbe/solver.py
```diff
@@ -960,9 +949,9 @@ def ao2mofn(moish):
     if rdm_return:
         rdm1 = make_rdm1_uccsd(ucc, relax=relax)
         if rdm2_return:
-            if use_cumulant:
-                with_dm1 = False
-            rdm2 = make_rdm2_uccsd(ucc, relax=relax, with_dm1=with_dm1)
+            rdm2 = make_rdm2_uccsd(
+                ucc, relax=relax, with_dm1=with_dm1 and not use_cumulant
```
lweisburn, line 953: This logic should be the same as above: `with_dm1` is True if `use_cumulant` is false, a separate `with_dm1` is unnecessary

src/quemb/molbe/lo.py
mcocdawc: that's an easy change. done

src/quemb/molbe/solver.py
mcocdawc: Note that `with_dm1` is also an argument to the function.

Before it was
```
if use_cumulant:
                with_dm1 = False
```
in particular it was **not**
```
with_dm1 = True
if use_cumulant:
                with_dm1 = False
```
as in other similar cases (which would allow your requested change to be fully equivalent)

My current change is exactly equivalent to the previous behaviour.


I can easily change to use only  `use_cumulant`, but then I will also remove the `with_dm1` argument

lweisburn: Yes, you're exactly right in the way you did this! I would just remove the `with_dm1` argument altogether, it's inherently redundant. You will never want the density generated with `with_dm1`True if you are running the code requesting the cumulant energy. Thank you for fixing/clarifying the redundancy in the current logic

mcocdawc: done :)

src/quemb/molbe/be_parallel.py
```diff
@@ -363,11 +343,8 @@ def run_solver_u(
...
     if ereturn:
-        if solver == "UCCSD" and not rdm_return:
-            with_dm1 = True
-            if use_cumulant:
-                with_dm1 = False
-            rdm2s = make_rdm2_uccsd(ucc, with_dm1=with_dm1)
+        if solver == "UCCSD" and not relax_density:
+            rdm2s = make_rdm2_uccsd(ucc, with_dm1=not use_cumulant)
         else:
             raise NotImplementedError("RDM Return not Implemented")
```
lweisburn, line 349: This is no longer the right error: I don't think we need an "else" error anymore. This should just be building the rdm2s when we have the unrelaxed density.

lweisburn: Everything looks good to me otherwise!!

lweisburn (approved):
It looks good to me!!

PR #77, opened 2024-12-23
Dataclass for solver kwargs

mcocdawc:
- Introduced a dataclass for solver kwargs. This is makes it much more explicit what arguments are supported by DMRG and SHCI

- Could simplify the interface of BE by removing a couple of arguments


lweisburn (approved):
This looks good to me! We (I) should work on setting up some SCI examples and tests to test out the code science-wise and think about the default parameters, but otherwise it all makes sense and makes the code more legible. We should confirm with @ShaunWeatherly to make sure the DMRG changes work for him

src/quemb/molbe/solver.py
```diff
@@ -123,21 +308,19 @@ def be_func(
             # pylint: disable-next=E0611
             from pyscf import hci  # noqa: PLC0415    # optional module
 
+            assert isinstance(solver_args, SHCI_ArgsUser)
+            SHCI_args = SHCI_Args.from_user_input(solver_args)
```
ShaunWeatherly, line 312: Should we keep the `A` in `SHCI_Args` consistently capitalized? Same with `DMRG_Args`?

src/quemb/molbe/solver.py
```diff
@@ -21,7 +24,187 @@
...
 from quemb.shared.manage_scratch import WorkDir
-from quemb.shared.typing import KwargDict
+
+
+class UserSolverArgs(ABC):
+    pass
+
+
+@define(frozen=True)
+class DMRG_ArgsUser(UserSolverArgs):
```
ShaunWeatherly, line 34: Maybe I'm missing something, but it seems like having a user interface class (DMRG_ArgsUser, SHCI_ArgsUser) which is distinct from the actual class used by the DMRG and SCI modules (DMRG_Args, SHCI_Args) adds a layer of ambiguity to the workflow. If there were other methods for constructing the `DMRG_Args` object _in addition_ to `from_user_input`, then I would see a reason for structuring things in this way. Is there a reason I'm not seeing for this particular organization of things?

This is especially salient when the only way we touch `DMRG_ArgsUser` is to convert it to the other class:

```
assert isinstance(solver_args, DMRG_ArgsUser)
DMRG_args = DMRG_Args.from_user_input(solver_args)
```

And to clarify what I mean by ambiguity, it looks like if I were to import `DMRG_Args` in the example DMRG script (instead of `DMRG_ArgsUser`) then we would run into an assertion error despite all of the same inputs being accepted. 

I guess this might also be resolved by allowing `DMRG_Args` to inherit `DMRG_ArgsUser`.

src/quemb/molbe/solver.py
```diff
@@ -21,7 +24,187 @@
 from quemb.shared.external.unrestricted_utils import make_uhf_obj
 from quemb.shared.helper import delete_multiple_files, unused
 from quemb.shared.manage_scratch import WorkDir
-from quemb.shared.typing import KwargDict
+
+
+class UserSolverArgs(ABC):
```
ShaunWeatherly, line 29: I feel like this might belong to `shared.typing`, although it would need a short docstring. I think it's general enough that it may find use in other modules (also it is used in both the pbe/molbe modules).

src/quemb/molbe/solver.py
```diff
@@ -21,7 +24,187 @@
 from quemb.shared.external.unrestricted_utils import make_uhf_obj
 from quemb.shared.helper import delete_multiple_files, unused
 from quemb.shared.manage_scratch import WorkDir
-from quemb.shared.typing import KwargDict
```
ShaunWeatherly, line 24: We are no longer importing `KwargDict` despite us still using them implicitly below, eg in `_get_schedule_kwargs_default`. Do we consider`KwargDict` deprecated and that's why?

src/quemb/molbe/solver.py
```diff
@@ -21,7 +24,187 @@
...
+    @classmethod
+    def from_user_input(cls, args: SHCI_ArgsUser):
+        if (args.select_cutoff is None) and (args.ci_coeff_cutoff is None):
+            select_cutoff = args.hci_cutoff
+            ci_coeff_cutoff = args.hci_cutoff
+        elif (args.select_cutoff is not None) and (args.ci_coeff_cutoff is not None):
+            ci_coeff_cutoff = args.ci_coeff_cutoff
+            select_cutoff = args.select_cutoff
+        else:
+            raise ValueError
```
ShaunWeatherly, line 200: Maybe something more descriptive:
```
raise ValueError(f" Solver args `ci_coeff_cutoff` and `select_cutoff` must both be specified or both be `None`!")
```

ShaunWeatherly (commented):
A few questions and potential changes, let me know your thoughts and if you agree then I'm happy to help make the suggested changes!

src/quemb/molbe/solver.py
mcocdawc: I tried to adhere to PEP8 while also accounting for capitalized acronyms such as DMRG or SHCI.
i.e. `SHCI_args` is an instance, while `SHCI_ArgsUser` is a class name.

I guess there is no right and wrong answer. :/

src/quemb/molbe/solver.py
mcocdawc: > My hunch is this, let me know if I'm wrong: you've moved away from using mutable defaults,

You got it right. :-)

A few more reasons on why I still think it is a good idea.

First, I really would not underestimate the advantage of immutable objects.

Yes, there is an additional layer of complexity, but that is not
there because I wanted to introduce an additional layer, but because it is intrinsically there.
In the previous version the variable `DMRG_solver_kwargs` was subtly changing
its meaning of "this is coming from user input" to, "it is now fully processed and initialized"
and can be used in `Block`.
Unfortunately it is not possible to initialize everything in the beginning, because there is a dependence of the defaults
on other state of the calculation such as `mf`.

You can see now just from the type if a variable is coming
from user input and still needs some processing to be a valid input to `Block`.
In addition, the interface of `from_user_input` makes it explicit that
there is a dependence on `mf`, which was somewhat under the radar otherwise.

> it looks like if I were to import DMRG_Args in the example DMRG script

good point
`DMRG_Args` is an implementation detail that the user should not import, I should name it `_DMRG_Args`.


The final processed `DMRG_Args` and `SHCI_Args` have no relation anymore and are independent types.

I introduced the abstract type only for the solver arguments to not bloat the interface of BE.
One could do this even without inheritance by introducing a union type `DMRG_ArgsUser | SHCI_ArgsUser`.

src/quemb/molbe/solver.py
mcocdawc: `KwargDict` is a dictionary mapping from `str` to `Any`. It is not deprecated, but in the particular case of `schedule_kwargs` we know that the values are either list of int or list of float, i.e. `dict[str, list[int] | list[float]]`

src/quemb/molbe/solver.py
mcocdawc: The final processed `DMRG_Args` and `SHCI_Args` have no relation anymore and are independent types.

I introduced the abstract type only for the user arguments to not bloat the interface of BE.
One could do this even without inheritance by introducing a union type `DMRG_ArgsUser | SHCI_ArgsUser`.

src/quemb/molbe/solver.py
mcocdawc: taken as is and done

src/quemb/molbe/solver.py
ShaunWeatherly: > DMRG_Args is an implementation detail that the user should not import, I should name it _DMRG_Args.

I would say this clears up the major point of confusion. The rest of my argument is mostly superficial then: a single class with two attributes and related class methods would be more readable _to me_ (ie. one component for immutable user input and another for fully initialized args). But again, with the ambiguity cleared up I don't think it's a real problem.

src/quemb/molbe/solver.py
ShaunWeatherly: AH I see that now, on first glance I just thought it may have been a typo. We should adhere to pep8, this is one of those stylistic standards that I actually really prefer.

ShaunWeatherly (approved):
Alright I think this looks good! The question about using `KwargDict` and `shared.typing` more generally still remains, but that's not so important as of right this instant.

Issue #80, opened 2025-01-08
Resolve `veff0` mismatch between molecular and periodic code


mscho527:
As discussed with @lweisburn in-person, these two arguments are necessary for now because of the mismatch in the `pFrag.veff0` attribute between the molecular and periodic code. This will (and should be) fixed in the future as a more fundamental fix, because we rely on this attribute for other projects. I will create an issue for this.

_Originally posted by @mscho527 in https://github.com/troyvvgroup/quemb/pull/79#discussion_r1907668030_

lweisburn:
Specifically, the object self.veff0 is None in in the molecular code right not, though it isn't in periodic. We should change this such that veff0 is usable

PR #81, opened 2025-01-09
Adding Veff0 to molecular code

lweisburn:
`veff0` (i.e. the `hf_veff` in each fragment Schmidt space), for each fragment, was removed from the molecular code at some point. This will be useful in the future for orbital optimization, etc, and was a discrepancy between the molecular and periodic versions. This pr adds back `veff0` to be constructed alongside `veff` (`fobj.veff0` and `fobj.veff` constructed during `cons_fock` in `molbe/pfrag.py`). This adds a return to the `get_veff` function in `molbe/helper.py`.

Adding `fobj.veff0` to match `pfobj.veff0`allows for the energy code to be simplified (getting rid of the pesky veff vs. veff0_par). We also no longer have to pass along `hf_veff` in `be_func` and `be_parallel` and recalculate `veff0` at every BE iteration.

(Addressing #80 )


mscho527:
Resolves #80

mscho527 (approved):
`veff0` and `veff` are confusing names (but are what we have inherited from the old code). we should address this when improving the naming scheme in a later PR.

```
⠀ ／l、          "approved!"
（ﾟ､ ｡ ７
⠀ l,‏‏‎ ‎‏‏‎ ‎‏‏‎ ‎‎~ヽ
  じしf_, )ノ
```

src/quemb/molbe/mbe.py
```diff
@@ -884,6 +883,8 @@ def initialize(self, eri_, compute_hf, restart=False) -> None:
             fobjs_.heff = numpy.zeros_like(fobjs_.h1)
             fobjs_.scf(fs=True, eri=eri)
 
+            # fobjs_.veff0 = multi_dot((fobjs_.TA.T, self.hf_veff, fobjs_.TA))
```
mcocdawc, line 886: we should aim to remove dead code. I will remove it in my PR.

lweisburn: Thanks! I was testing something and forgot to remove this:)

PR #82, opened 2025-01-09
Simplify init

mcocdawc:


mscho527:
I'll have to think a little bit more, but I am a bit wary of directly using `self.fobj`. If we are concerned about the redundant information we are retaining here, is there a reason why we wouldn't take the information from the fobj in the `init` and not retain `self.fobj` instead? I'll think about this a bit longer and update lol; just my very first thoughts :)

mscho527 (approved):
cool cool nice

PR #83, opened 2025-01-10
Make numpy shorter

mcocdawc:
- replacing `np.dot` with `@`
- replacing the `numpy.something` calls with `from numpy import something` if functions appear several times
- replacing the `numpy.something` call with `np.something` if a function appears rarely

### DIFF
--- a/src/quemb/kbe/pfrag.py
+++ b/src/quemb/kbe/pfrag.py
@@ -2,7 +2,18 @@
 
 
 import h5py
-import numpy
+from numpy import (
+    abs,
+    complex128,
+    diag_indices,
+    einsum,
+    outer,
+    result_type,
+    trace,
+    tril_indices,
+    zeros,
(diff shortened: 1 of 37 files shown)


src/quemb/kbe/pfrag.py
```diff
@@ -161,12 +162,10 @@ def sd(
         supcell_rdm = einsum("Rk,kuv,Sk->RuSv", phase, rdm1_lo_k, phase.conj())
         supcell_rdm = supcell_rdm.reshape(nk * nlo, nk * nlo)
 
-        if abs(supcell_rdm.imag).max() < 1.0e-6:
+        if (max_val := np.abs(supcell_rdm.imag).max()) < 1.0e-6:
```
mcocdawc, line 165: first time walrus operator

mscho527 (approved):
cool nice cool!

PR #85, opened 2025-01-13
added chemical fragmentation

mcocdawc:
- use van der Waals radii (relies on chemcoord)
- much nicer code

will make heavy use of the tests introduced in #86 ; thanks @ShaunWeatherly


mcocdawc:
- [ ] time chemfrag, graphgen and autogen

mcocdawc:
Close and do it cleanly in #99 instead

PR #86, opened 2025-01-14
Graph theoretic fragmentation via `graphgen`.

ShaunWeatherly:
The idea is rather straightforward: 

`graphgen` generalizes the BE _n_ fragmentation scheme to arbitrary fragment sizes (I've tested BE1-BE9 so far) by using a graph theoretic heuristic. In it, atoms are assigned to nodes in an adjacency graph and edges are weighted by some distance metric. For a given fragment center site, Dijkstra's algorithm is used to find the shortest path from that center to its neighbors. The number of nodes visited on that shortest path determines the degree of separation of the corresponding neighbor. I.e., all atoms whose shortest paths from the center site visit at most 1 node must be direct neighbors (adjacent to the center site), which gives BE2-type fragments; all atoms whose shortest paths visit at most 2 nodes must then be second-order neighbors, hence BE3; and so on. This depends on the [NetworkX library](https://networkx.org/documentation/stable/reference/introduction.html).

Major points to note:
1. Works for arbitrary _n_ in BE _n_. (Yields _near_ identical fragments to `autogen` for n=1->3) 
2. Does **not** rely on heuristic bond lengths or atom identities.
3. _Much simpler code_, which I think we can all appreciate.
4. And most importantly - the distance metric for assigning edge weights in the adjacency graph can easily be changed. For example, I'm currently looking into fragmentation via entanglement (orbital pair mutual info).

EXTRA: now adds `FragmentMap` data class.
EXTRA2: there are 14 new unittests covering both `autogen` and `graphgen`


src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
-# Author: Oinam Romesh Meitei
-
+# Author: Oinam Romesh Meitei, Shaun Weatherly
 
+import networkx as nx  # type: ignore
```
mcocdawc, line 3: Required:
1. In general I would always ignore specific errors, namely `ignore-missing-imports`.
2. In general I would do a package-widge `ignore-missing-imports` in the `mypy.ini` file in the root of the repo. 
3. In this particular case there is a types package https://pypi.org/project/types-networkx/. I would add it to the [static_analysis_requirements.txt](https://github.com/troyvvgroup/quemb/blob/main/tests/static_analysis_requirements.txt) file

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
 
 from quemb.molbe.helper import get_core
 from quemb.shared.helper import unused
 
 
+def euclidean_norm(
+    i_coord: float,
+    j_coord: float,
+):
+    return np.linalg.norm(np.asarray(i_coord - j_coord))
```
mcocdawc, line 16: Required:
Either `np.linalg.norm` or `norm` should be used consistently.

(Personally I prefer `norm`)

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+from pyscf import gto
 
 from quemb.molbe.helper import get_core
 from quemb.shared.helper import unused
 
 
+def euclidean_norm(
+    i_coord: float,
+    j_coord: float,
+):
```
mcocdawc, line 15: I don't want to force type-hinting, but if you use it, then I would also add the return type.
i.e. `-> float` in this case.

This comment applies to the rest of the code

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+    from that center to its neighbors. The number of nodes visited on that shortest
+    path determines the degree of separation of the corresponding neighbor. I.e.,
+    all atoms whose shortest paths from the center site visit at most 1 node must
+    be direct neighbors to the center site, which gives BE2-type fragments; all
+    atoms whose shortest paths visit at most 2 nodes must then be second-order
+    neighbors, hence BE3; and so on.
+
+    Parameters
+    ----------
+    mol : pyscf.gto.mole.Mole
```
mcocdawc, line 61: the types in the parameters are redundant, since sphinx takes it from the type annotation.

I would remove them here

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+
+
+def graphgen(
+    mol: gto.Mole,
+    be_type: str = "BE2",
+    frozen_core: bool = True,
+    remove_nonunique_frags: bool = True,
+    frag_prefix: str = "f",
+    connectivity: str = "euclidean",
+    # draw_graph: bool = True,
```
mcocdawc, line 44: I would remove commented code

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
 
 
+def euclidean_norm(
+    i_coord: float,
+    j_coord: float,
+):
+    return np.linalg.norm(np.asarray(i_coord - j_coord))
+
+
+def remove_nonnunique_frags(
```
mcocdawc, line 19: I would not change fragment_map in place. It seems to be a cheap data structure.

If you want to change it in place, then I would return None.

(Note that not specifying a return type is equivalent to `-> Any`, hence you would need to specify `-> None` in this case)

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+        adx: {
+            "bas": bas,
+            "label": mol.atom_symbol(adx),
+            "coord": mol.atom_coord(adx),
+            "shortest_paths": dict(),
+        }
+        for adx, bas in enumerate(mol.aoslice_by_atom())
+    }
+
+    fragment_map = {
```
mcocdawc, line 97: If you declare a FragmentMap Type alias outside the function, you can use it here **and** you can improve the interface of e.g. `remove_nonnunique_frags` 
by specifying more than just `dict` as type (which is equivalent to `dict[Any, Any]`)


Furthermore, defining a dataclass is so cracy simple with `attrs.define` that I would actually do that instead of a dictionary. This makes it far easier to define the expected entries and makes it impossible to use wrong keys.

Let's better talk in person about this. :-)

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+            )
+
+            # If the degree of separation is smaller than the *n*
+            # in your fragment type, BE*n*, then that site is appended to
+            # the set of fragment sites for adx.
+            for bdx, path in map["shortest_paths"].items():
+                if 0 < (len(path[0]) - 1) < fragment_type_order:
+                    fsites_temp = fsites_temp + fragment_map["sites"][bdx]
+                    fs_temp.append(fragment_map["sites"][bdx])
+
```
mcocdawc, line 168: is it really necessary to make all these explicit casts to a tuple, or could you just declare the entries a `Sequence` (i.e. non-modifyable list-like object)
this goes hand in hand with my other comment about using a `attrs.define` dataclass.

src/quemb/molbe/autofrag.py
```diff
@@ -1,12 +1,221 @@
...
+        raise AttributeError(f"Connectivity metric not recognized: '{connectivity}'")
+
+    # Remove all fragments whose AO indices can be identified as subsets of
+    # another fragment's. The center site for the removed frag is then
+    # added to that of the superset. Because doing so will necessarily
+    # change the definition of fragments, we repeat it up to `natm` times
+    # such that all fragments are guaranteed to be distinct sets.
+    if remove_nonunique_frags:
+        for _ in range(0, natm):
+            fragment_map = remove_nonnunique_frags(fragment_map)
```
mcocdawc, line 189: here you use it as if `remove_nonunique_frags` was side-effect-free, but it isn't. :-(

src/quemb/molbe/fragment.py
```diff
@@ -82,24 +82,47 @@ def __init__(
...
+                frozen_core=frozen_core,
+                remove_nonunique_frags=True,
+                frag_prefix="f",
+                connectivity="euclidean",
+            )
+
+            self.fsites = fragment_map["fsites"]
+            self.edge = fragment_map["edge"]
+            self.center = fragment_map["center"]
+            # self.edge_idx = fragment_map["edge"]
```
mcocdawc, line 116: dead code should be removed

mcocdawc (changes requested):
Great improvement! Thank you!

I have a few strongly requested changes and some other comments.

mcocdawc:
oh and the most important comment, I really would add a test, probably something like `molbe_octane_test.py` just with this fragmentation. this could ensure, that they indeed produce the same results for BE1-3

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
ShaunWeatherly: I'll push back a bit on this one, there's nothing inherently wrong with lines of code which have been commented out. In this case I agree it should be removed, but in general I would disagree. These are harmless bits of extra info that people learning the codebase and making their first contributions would appreciate.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done, but I don't really see the point in this if the data class is only used by `graphgen`. If you plan to expand on it then I can see the value in it, but as it is it kinda just adds bloat to the module.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Nope. But there's no harm in it.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Fixed!

src/quemb/molbe/fragment.py
ShaunWeatherly: Here in particular - this is an attribute that is used in autofrag, but not used in my code. If we plan to remove `self.edge_idx` it in the future, then that would justify removing the commented line of code. But as is it does no harm and simply helpes with bookkeeping/making comparisons to autofrag/etc.

ShaunWeatherly:
Alright, unittests have been added for both `autogen` and `graphgen`, likewise I think I've addressed all of your other comments, but do let me know if there's anything I've missed!

src/quemb/molbe/autofrag.py
```diff
@@ -1,18 +1,291 @@
...
+    else:
+        raise AttributeError(f"Connectivity metric not recognized: '{connectivity}'")
+
+    # Remove all fragments whose AO indices can be identified as subsets of
+    # another fragment's. The center site for the removed frag is then
+    # added to that of the superset. Because doing so will necessarily
+    # change the definition of fragments, we repeat it up to `natm` times
+    # such that all fragments are guaranteed to be distinct sets.
+    if remove_nonunique_frags:
+        for _ in range(0, natm):
```
mcocdawc, line 250: can this outer loop not be a part of 
`remove_nonnunique_frags`?

mcocdawc (commented):
Looks really good now.
The only big question mark I still have is about the `euclidean_norm` function.

A general comment just for the protocol and potential improvements for the future; we talked about it in person.
I think there is unnecessary casting between lists, tuples and sets.
This graph operation can most likely be completely done with `sets` and `frozenset` with only one cast to an ordered container such as list in the end.
In the current implementation there is more book-keeping necessary to avoid duplicates and unnecessary frequent casts.
But the code works, is really well tested, and readable. If you want to keep it like this, then I am also fine.

.ruff.toml
```diff
@@ -1,6 +1,7 @@
 [lint]
 # see the rules [here](https://docs.astral.sh/ruff/rules/)
 select = ["E", "F", "I", "NPY", "PL", "ARG"]
+exclude = ["tests/fragmentation_test.py"]
```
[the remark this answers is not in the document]
ShaunWeatherly: Yep, this is remnant of something I already fixed and forgot to remove.

src/quemb/molbe/autofrag.py
mcocdawc: I also agree with you. Sometimes we want to keep it.
I should have written:
"I would remove *this* commented code" :D

(nevertheless we use `git`, any deleted code is still there if we really need it. So in doubt I am leaning towards deleting.)

.ruff.toml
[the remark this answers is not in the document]
mcocdawc: just for reference

> You can also disable all errors of a given kind across an entire file like so:

```
# ruff: noqa: F401
```

src/quemb/molbe/autofrag.py
```diff
@@ -1,18 +1,291 @@
-# Author: Oinam Romesh Meitei
+# Author: Oinam Romesh Meitei, Shaun Weatherly
 
+from typing import Any
 
+import networkx as nx
+import numpy as np
+from attrs import define
+from networkx import single_source_all_shortest_paths  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
ShaunWeatherly: I think this is a missing attribute in the extra `networkx` typing library. Can be removed in the future.

tests/fragmentation_test.py
```diff
@@ -0,0 +1,1419 @@
...
+        mol = gto.M()
+        mol.atom = [["H", (0.0, 0.0, i)] for i in range(8)]
+        mol.basis = "sto-3g"
+        mol.charge = 0.0
+        mol.spin = 0.0
+        mol.build()
+
+        mf = scf.RHF(mol)
+
+        target = {
```
mcocdawc, line 26: Sorry a last comment which I just realized now.

It would make it much easier to see where you actually expect the same result for `autogen`  and `graphgen` if there is a loop for those systems where it should be the same.

This would be a great improvement also for future code. (I could then for example just extend the loop)

i.e. a function
```python3
def test_h_linear_be1(...):
    mol, mf, .... = prepare_system
    target = ...
    for frag_method in ["graphen", "autogen"]:
           calculated = ...
           assert calculated == target
```

ShaunWeatherly: I would agree, except we don't expect `autogen` and `graphgen` to yield exactly the same indexing even for small fragment sizes. These tests ensure the indexing is internally consistent within each frag type.

src/quemb/molbe/lo.py
```diff
@@ -360,7 +362,7 @@ def localize(
                 W_ = C_ @ W_
 
             self.W = get_loc(
-                self.mol, W_, "PM", pop_method=pop_method, init_guess=init_guess
+                self.mf.mol, W_, "PM", pop_method=pop_method, init_guess=init_guess
```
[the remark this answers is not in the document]
ShaunWeatherly: BE.mol does not exist anymore, likely because someone removed it at some point.

src/quemb/molbe/autofrag.py
```diff
@@ -1,18 +1,291 @@
...
+
+    # Define the 'edges' for fragment A as the intersect of its sites
+    # with the set of all center sites outside of A:
+    for adx, fs in enumerate(fragment_map.fs):
+        edge: set[tuple] = set()
+        for bdx, center in enumerate(fragment_map.center):
+            if adx == bdx:
+                pass
+            else:
+                overlap = set(fs).intersection(set((center,)))
```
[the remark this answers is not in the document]
ShaunWeatherly: As it is written is much more readable in my opinion. I vote to keep it.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Sure. Just note that I'm going to request _at least_ 20 changes on your next PR as well.

tests/fragmentation_test.py
```diff
@@ -0,0 +1,1419 @@
+"""
```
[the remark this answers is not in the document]
ShaunWeatherly: Done.

src/quemb/molbe/solver.py
```diff
@@ -908,10 +908,10 @@ def solve_block2(
     mc.fcisolver.twodot_to_onedot = DMRG_args.twodot_to_onedot
     mc.fcisolver.maxIter = DMRG_args.max_iter
     mc.fcisolver.block_extra_keyword = DMRG_args.block_extra_keyword
-    mc.fcisolver.scratchDirectory = str(frag_scratch)
-    mc.fcisolver.runtimeDir = str(frag_scratch)
+    mc.fcisolver.scratchDirectory = frag_scratch.path
+    mc.fcisolver.runtimeDir = frag_scratch.path
     mc.fcisolver.memory = DMRG_args.max_mem
-    os.chdir(frag_scratch)
+    os.chdir(frag_scratch.path)
```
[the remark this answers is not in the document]
ShaunWeatherly: In the case you highlight here yes, you'd be correct, but for `mc.fcisolver.scratch` above this does not work, and a string is explicitly required.

src/quemb/molbe/autofrag.py
```diff
@@ -1,18 +1,291 @@
...
+from attrs import define
+from networkx import single_source_all_shortest_paths  # type: ignore[attr-defined]
 from numpy.linalg import norm
+from pyscf import gto
 
 from quemb.molbe.helper import get_core
 from quemb.shared.helper import unused
 
 
+@define(init=True)
```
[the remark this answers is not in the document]
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
```diff
@@ -1,18 +1,291 @@
...
+                    tmp = set(self.center[adx] + self.center[bdx])
+                    self.center[adx] = tuple(tmp)
+                    del self.center[bdx]
+                    del self.fsites[bdx]
+                    del self.fs[bdx]
+
+        return None
+
+
+def euclidean_norm(
```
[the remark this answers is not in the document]
ShaunWeatherly: > If i_coord and j_coord are actually vectors, then the types are wrong, and still it is the euclidean metric (not the norm).

They are indeed supposed to be vectors, that's a good catch, but the second half of this statement is wrong. Any normed space is a metric space. You could make an argument that `norm` is slightly slower than directly computing the squares, sums, and roots, but that'd really be getting into over-optimization territory.

src/quemb/molbe/autofrag.py
mcocdawc: looking forward :-)

src/quemb/molbe/autofrag.py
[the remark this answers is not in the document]
mcocdawc: > Any normed space is a metric space

Indeed, and any norm induces a metric. **But** that still does not mean, that a norm and a metric are the same.
A norm is a function of one vector and a metric is a function of two vectors (plus some other properties of course).
This function is not a norm, but a metric.

My argument is not about performance, but because of the `norm` in the name and declaration of two floats I was genuinely confused. 
Reading the function I saw two possibilities; maybe you meant the euclidean metric/distance for arbitrary dimensions:
```python3
def euclidean_metric(a: Vector, b: Vector) -> float:
    return norm(a - b)
```
or perhaps you indeed meant the norm for a 2-dimensional space
```python3
def euclidean_norm(x: float, y: float) -> float:
    return np.sqrt(x**2 + y**2)
```

Since I did not know yet your algorithm, I imagined that perhaps a projection onto a 2D-plane happens and you need to calculate the norm/length on this plane so often that you introduce this function.
Reading on I realized that this is not the case, **but** this confusion can be easily averted by giving this function any of the following names:
- `euclidean_metric`
- `euclidean_distance`
- `metric`
- `distance`
- `get_distance`
- ....

just not norm, because it really is not a norm.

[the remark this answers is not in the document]
ShaunWeatherly: It is the norm of the vector separating the two atoms in real space. That the _python function_  takes two inputs does not change this, as what we are computing here is the _norm_ of the vector which separates the two atoms. This vector defines the euclidean distance, and its norm is the euclidean norm. It is a norm.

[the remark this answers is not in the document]
mcocdawc: You calculate the **distance** between two atoms, that is a metric.
This **euclidean metric** is induced by the **euclidean norm**.

We can even test the defining axioms for the [norm](https://en.wikipedia.org/wiki/Norm_(mathematics)) and [metric](https://en.wikipedia.org/wiki/Metric_space) .
Assuming `x, y, z` are random vectors of same dimension, and assuming that your function is called `euclidean_metric` we have
```python3
assert euclidean_metric(x, x) == 0
assert euclidean_metric(x, y) == euclidean_metric(y, x)
assert euclidean_metric(x, z) <=  euclidean_metric(y, z) + euclidean_metric(y, z)
```

While, as expected, `numpy.linalg.norm` is a norm. (assuming that `a` is scalar)
```python3
from numpy.linalg import norm

assert norm(x + y) <= norm(x) + norm(y)
assert norm(a * x) <= abs(a) * norm(x) 
assert (norm(x) == 0) == (all(x == 0))
```

calling it `norm` while there is a well-defined correct name is wrong and misleading.

mcocdawc (approved):

PR #91, opened 2025-01-20
Indexing fixes, optimization, and etc. for `graphgen`

ShaunWeatherly:
A few quick fixes. Perhaps next time just send me a message on slack, @mcocdawc .

### Incomplete Initialization
- `Frag_atom`, along with all of the additional attributes generated by `autogen` and not generated by `graphgen`, are currently unused in the code. As per _your_ insistence on removing "dead code", I excluded the initialization of these parameters, because it was, and is, unused. I don't see how this could be described as "incomplete initialization", but I'll just assume that you plan to make use of this attribute and I've added it back in at your request.

### Fails at Molecule
- This is an edge case, but it is a good catch. There was an issue with `remove_nonnunique_frags` that messed up the indexing when `Natm` was very large. This led to a number of smaller issues, along with the issue you encountered of some center sites not being appended to the superset fragments. I've fixed the issue and everything seems to be working just fine on my end now, but let me know if you encounter any other abnormalities.

### Bad Scaling
- Yup, that was expected. `graphgen` previously found the shortest paths from a center to _all_ other atoms in the system. Obviousuly, finding the shortest path from all-to-all is non-trivial for graphs with 100s of nodes. The fix is straightforward: 1) exclude edges from `adjacency_graph` if they are below some threshold, and 2) only find the shortest path for neighboring atoms which have an edge weight above that threshold. I've implemented that fix and the threshold can be found as `graphgen.cutoff`. 

### Misleading name and interface
- Okay, how about this. @mscho527 @lweisburn , if you agree that this function is terribly named, misleading, and must be changed, then I'll change it. If not, then it stays. I prefer this name and I don't find it misleading in the slightest, personally, and frankly I think this type of pedantry is a bit counter productive. But again, I'm open to admitting wrongness if others also find it misleading.


mcocdawc:
> graphgen previously found the shortest paths from a center to all other atoms in the system.

Ok, so it went from N^2 scaling to linear one? 

> As per your insistence on removing "dead code", I excluded the initialization of these parameters, because it was, and is, unused.

I softly suggested to remove **commented out code**, if you had wanted to keep it it would have been fine as well.
In general I would also remove locally unused code, **but** values that are returned via a public API cannot known to be unused. I did not and would not suggest to remove them.
Admittedly I missed in my review that there are values missing from the returned code and will pay closer attention to this case in the future.

> if you agree that this function is terribly named

I never said it is terribly named. Asking it in this way is a bit of a rhetorical question.
Metric and norm are clearly defined and different mathematical terms.
Using one for the other is misleading.
In the same way as a variable `hermitian_M` for a matrix that is actually unitary, or a variable name `i_AO_idx` for an MO index is subtly wrong and dangerously misleading.

ShaunWeatherly:
@mcocdawc
1) That should be order _constant_ scaling per fragment (or _linear_ in the number of fragments), just like autogen I believe.
2) 
3) 

Again, every norm induces a metric, but not every metric induces a norm. Norms have distinct properties, such as translational invariance. Metrics are more general, and the use of `norm` here agrees with the more strict properties of a norm. To say the only difference is that "a metric is a function of two vectors whereas a norm is a function of one" is an oversimplification, because you can always just define the origin to be one of the two coordinates, ie. ||x-y|| = d(x,y), such that d(u,0) is both a norm and a metric. 

But semantics aside, the function `eucidean_norm` _satisfies your own tests for defining a norm_. To recall, you gave the following properties as those defining a metric:
```
assert euclidean_metric(x, x) == 0
assert euclidean_metric(x, y) == euclidean_metric(y, x)
assert euclidean_metric(x, z) <=  euclidean_metric(y, z) + euclidean_metric(y, z)
```
whereas a norm would satisfy:
```
assert norm(x + y) <= norm(x) + norm(y)
assert norm(a * x) <= abs(a) * norm(x) 
assert (norm(x) == 0) == (all(x == 0))
```
Naturally, `euclidean_norm` satisfies the properties of a metric, because a norm induces a metric and hence it would satisfy those properties regardless of whether it were a norm or a metric. What's meaningful is if you test `euclidean_norm` for the properties of a norm:
```

a = np.random.rand(1, )[0]
x1 = np.random.rand(3, 1)
x2 = np.random.rand(3, 1)
y1 = np.random.rand(3, 1)
y2 = np.random.rand(3, 1)

assert euclidean_norm(x1 + x2, y1 + y2) <= euclidean_norm(x1, y1) + euclidean_norm(x2, y2)
assert euclidean_norm(a * x1, a * y1) <= abs(a) * euclidean_norm(x1, y1) 
assert (euclidean_norm(x1, y1) == 0) == (all(x1 == 0))

```
You find that _by your own tests, it satisfies the properties of a norm_. I truly don't understand your insistence on this, ultimately trivial, point, but if you insist on brick walling then feel free to change the name of the function to whatever you like.

mcocdawc:
> Then I think you need to make your standards for this code more clear, because it's currently impossible to tell what you'll find issue with.

I don't define the standards for the code alone.
There were only two problems where I explicitly required a change, which I indicated by `required` in the title.
The misleading naming of `euclidean_norm` and [this here](https://github.com/troyvvgroup/quemb/pull/86#discussion_r1915500622)

All other comments were just comments of the kind "I would do it like this". 
If you agree, great, if you don't agree it is also fine.

--------------------------------------------------------------------
If I see a function call `euclidean_norm(x, y)`, then I assume that it calculates `sqrt(x**2 + y**2)`, i.e. the euclidean norm of a 2D vector.
If I see a function call `euclidean_metric(x, y)`, then I assume that it calculates `euclidean_norm(x - y)`. 

I think that others also would make the same assumption, hence would get confused by the current naming. So let's just wait for the comment of @mscho527 and @lweisburn , perhaps I am wrong then it stays as it is.

mscho527:
@ShaunWeatherly and @mcocdawc, it is a bit difficult for me to exactly follow everything since i) multiple issues are discussed at the same time and ii) from what I understand, is based off of what you discussed in-person between you two. Do you want to walk us through in-person? Perhaps during the subgroup tomorrow?

src/quemb/molbe/autofrag.py
```diff
@@ -1,16 +1,18 @@
...
 
-from typing import Any
+from copy import deepcopy
+from typing import Sequence
 
 import networkx as nx
 import numpy as np
 from attrs import define
-from networkx import single_source_all_shortest_paths  # type: ignore[attr-defined]
+from networkx import shortest_path  # type: ignore[attr-defined]
```
mcocdawc, line 9: is the `type ignore` still necessary here after the change?

src/quemb/molbe/autofrag.py
```diff
@@ -75,25 +85,42 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
 
         return None
 
 
-def euclidean_norm(
-    i_coord: np.ndarray,
-    j_coord: np.ndarray,
-) -> np.floating[Any]:
-    return norm(np.asarray(i_coord - j_coord))
+def euclidean_distance(
```
mcocdawc, line 119: what a pleasant surprise :)


I think with the new interface the `def ...` become one line.

src/quemb/molbe/autofrag.py
```diff
@@ -260,22 +304,26 @@ def graphgen(
...
-                if overlap:
-                    edge = edge.union(overlap)
-        fragment_map.edge.append(tuple(edge))
+                for f in fs:
+                    overlap = set(f).intersection(set(center))
+                    if overlap:
+                        f_temp = set(fragment_map.fragment_atoms[adx])
+                        c_temp = set(fragment_map.center_atoms[bdx])
+                        edge_temp.add(tuple(overlap))
+                        eatoms_temp.add(i for i in f_temp.intersection(c_temp))
```
mcocdawc, line 319: do you really want to add an iterator as element to the set, or do you want to add the elements `i` to the set?

Currently it is the iterator, which is a bit surprising. 

If this is really the desired behaviour then I would declare `eatoms_tmp` as list, since you don't make any use of set - behaviour.

```python3
{(i for i in [1, 2]), (i for i in [1, 2])}
```
yields 
```python3
{<generator object <genexpr> at 0x7d803e219cc0>,
 <generator object <genexpr> at 0x7d803e21ab00>}
```

i.e. they are considered different generator expressions.

src/quemb/molbe/autofrag.py
```diff
@@ -260,22 +304,26 @@ def graphgen(
...
+        fragment_map.edge.append(tuple(edge_temp))
+        fragment_map.edge_atoms.append(tuple(eatoms_temp))
 
     # Update relative center site indices (centerf_idx) and weights
     # for center site contributions to the energy (ebe_weights):
     for adx, center in enumerate(fragment_map.center):
-        centerf_idx = tuple(
-            set([fragment_map.fsites[adx].index(cdx) for cdx in center])
-        )
+        centerf_idx = tuple([fragment_map.fsites[adx].index(cdx) for cdx in center])
```
mcocdawc, line 326: if you cast to a tuple you can omit the list expression and use the generator expression instead. 

```python3
tuple(fragment_map.fsites[adx].index(cdx) for cdx in center)
```

src/quemb/molbe/fragment.py
```diff
@@ -70,9 +90,13 @@ def __init__(
         self.center_idx = []
         self.centerf_idx = []
         self.be_type = be_type
+        self.frag_prefix = frag_prefix
+        self.connectivity = connectivity
         self.frozen_core = frozen_core
         self.iao_valence_basis = iao_valence_basis
         self.valence_only = valence_only
+        self.cutoff = cutoff
+        self.remove_nonnunique_frags = remove_nonnunique_frags
```
mcocdawc, line 99: From my understanding the `frag_prefix`, `connectivity`, `cutoff`, `remove_nonnunique_frags`
are only arguments for `graphgen`.

This has two implications.
1. They should not be components of `self`. Otherwise other codes also have to support them, even if it does not make sense there.
2. I would actually collect these implementation specific arguments either into a dictionary or preferably a dataclass similar to the DMRG or SHCI args.

tests/fragmentation_test.py
```diff
@@ -1389,9 +1231,9 @@ def run_energies_test(
                     float(E_B),
                     msg=f"{test_name}: BE Correlation Energy (oneshot) for "
                     + frag_type_A
```
mcocdawc, line 1233: this could become just one f-string without mixing old-school + and f-strings

tests/fragmentation_test.py
```diff
@@ -1405,13 +1247,19 @@ def run_indices_test(
     ):
         fobj = fragpart(frag_type=frag_type, be_type=be_type, mol=mf.mol)
         try:
+            # For debugging:
```
mcocdawc, line 1250: I would remove these comments.

mcocdawc (commented):
Looks nice overall :D

I still see an inconsistency between `autogen` and `graphgen` in the meaning of a couple of `fragpart` attributes, e.g. `self.center`. But
1. This PR definitely fixes bugs and makes it better
2. I still try to understand myself where this inconsistency comes from.

We should discuss this on the blackboard.

src/quemb/molbe/autofrag.py
ShaunWeatherly: No longer necessary!

src/quemb/molbe/autofrag.py
ShaunWeatherly: Done.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Good catch! That should be addition of elements to the set.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Yep, done.

src/quemb/molbe/fragment.py
ShaunWeatherly: The same could be said for the mixed-basis BE attributes:
```
        # Initialize class attributes necessary for mixed-basis BE
        self.Frag_atom = []
        self.center_atom = []
        self.hlist_atom = []
        self.add_center_atom = []
```
along with a few other attributes. If we want to change the data structure of `fragpart`, that should probably be a separate PR.

tests/fragmentation_test.py
ShaunWeatherly: Eh this was a quick fix for the line length limit. I can change it if you really want, but it's ultimately superficial.

tests/fragmentation_test.py
ShaunWeatherly: Sure.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Ah actually, `ruff` prefers what it was originally. I'll revert this.

src/quemb/molbe/autofrag.py
mcocdawc: Great to hear that. :D
The reason why I got stuck on it, was that `eatoms_temp` had not any type for its elements declared and I tried to figure the correct one out. 

If it is declared as `set[int]`, then I think `mypy` should catch this error automatically.

src/quemb/molbe/fragment.py
mcocdawc: I just talked to @lweisburn and she agrees with removing `self.hlist_atom` and `self.add_center_atom` . (But as you said, to be done in a separate PR)

On the other hand we both think that `self.Frag_atom`, `self.center_atom`  is so important and useful, that it actually should be exported by every fragmentation code. (In particular it is defined for every fragmentation code.)

src/quemb/molbe/autofrag.py
```diff
@@ -260,22 +304,26 @@ def graphgen(
     # Define the 'edges' for fragment A as the intersect of its sites
     # with the set of all center sites outside of A:
     for adx, fs in enumerate(fragment_map.fs):
-        edge: set[tuple] = set()
+        edge_temp: set[tuple] = set()
+        eatoms_temp: set = set()
```
mcocdawc, line 308: (I don't know if you get a notification for resolved conversations, so I repeat it here)

If `eatoms_temp` would have been declared as `set[tuple]` (or even better `set[tuple[int, ...]]`) then the generator expression error would have been caught by `mypy`.

I really would add this declaration.

mcocdawc (approved):
Looks really good now.

Please note my comment about `eatoms_temp` where I would still add the type declaration.

src/quemb/molbe/autofrag.py
ShaunWeatherly: Got it, I've added the declaration.

PR #94, opened 2025-01-22
Remove requirement for scratch directory to be empty: works with mixe…

lweisburn:
Removes the requirement for the scratch directory to be empty. This allows for mixed-basis BE, which writes scratch to the same directory, to work with QuEmb. I don't think this otherwise would be false (and even if it is, I don't see when it would be an issue for the scratch to be nonempty)


mcocdawc:
In the docstring of `WorkDir` it says
>     If it already exists, it is ensured, that it is empty.

This has to be accordingly removed as well.

Otherwise it looks good (y)

mcocdawc (changes requested):
In the docstring of `WorkDir` it says
>     If it already exists, it is ensured, that it is empty.

This has to be accordingly removed as well.

Otherwise it looks good (y)

mcocdawc (approved):
Cool cool :D

PR #96, opened 2025-01-22
Fixes for DMRG test suite and related bugs.

ShaunWeatherly:
**Major Changes:**
 - Modifies the tests in `dmrg_molbe_test` to run in github actions. 
 - Exposes useful scratch management options to high level modules, ie`mbe.BE`. Namely: one can now specify `scratch_dir` and `cleanup_at_end` when instantiating the `BE` object.
 - Adds missing DMRG modules to `be_parallel`.

**Specifics:**
 - Fixes to `mbe.BE.scratch_dir`. This keyword was broken by the recent changes to scratch management - now it is correctly passed to `WorkDir` and used to generate the scratch root.
 - Improved caching for different parts of the code in `quemb_unittest.yml`. Currently separated into `pip-cache`, `quemb-cache` and `block2-cache`. Also makes cache keys more descriptive, which fixes a bug where caches failed to update despite dependencies changing.
 - DMRG scratch cleanup now relies on `WorkDir.cleanup_at_end`, and the old kwarg (`force_cleanup`) has been removed.
 - Fixes reference to `rets[0]` as the total BE energy in `solver`, which is incorrect.

Full commit history ~

* Fixes for scratch management in `solver` and `helper` modules.

* Added `DMRG_ArgsUser` to `TestBE_DMRG`

* Fixed incorrect definition of `ebe_tot` as `rets[0]`, which is only the correlation energy.

* Make `helper.delete_multiple_files` recursive.

* Make `use_cumulant=True` by default in `solver_block2`.

* Modify `molecular_DMRG_test`

* Remove block2 versioning

* Test explicit cache save/restore.

* Fully explicit save/restore for cache

* Block2 `no-binary`

* Test `no-binary`

* Typo

* Explicit block2 versioning

* Move to `test_requirements`

* Edit block2 versioning.

* Force latest `mkl`

* test latest `mkl`

* Separate caching for `block2`

* Test caching

* Test caching

* Fine tune caching

* Further fine-tune caching

* Back to block2 versioning

* Formatting and `ompnum` passing.

* Install quemb in workflows.

* resolve `molecular_DMRG_test` scratch

* Update example script

* Update .gitignore

* Enable all unittests

---------


ShaunWeatherly:
Note to @lweisburn and @mscho527 , this PR corrects the definition of `BE.ebe_tot` to be the total BE energy. Previously, some parts of the code were using it as the correlation energy, including in the `misc.be2puffin` function. I've updated that function to explicitly use the correlation energy (which I assume was the original intent) and I just wanted to double check and make sure I didn't break anything!

mscho527:
I haven't read the details, but this seems like a substantial change in the code with a single PR. Shall we ask 2+ approvals as we have done for previous large PRs? Also, it will take a bit for me to parse through this.

src/quemb/molbe/solver.py
```diff
@@ -908,35 +916,57 @@ def solve_block2(
...
-    mc.kernel(orbs)
-    rdm1, rdm2 = dmrgscf.DMRGCI.make_rdm12(
-        mc.fcisolver, DMRG_args.root, DMRG_args.norb, DMRG_args.nelec
-    )
+    if DMRG_args.force_earlystop:
+        # Generates the input files for block2 but stops short of calling it.
+        # Instead loads the 1- and 2-RDM from `scratch` and proceed with BE.
+        # (Currently this is only used in the unit tests: `dmrg_molBE_test`)
+        dmrgscf.dmrgci.writeDMRGConfFile(mc.fcisolver, DMRG_args.nelec, Restart=False)
+        rdm1_path = Path(scratch + "/rdm1.npy")
```
mcocdawc, line 928: `Path(scratch + "/rdm1.npy")`
is much nicer written as 
`frag_scratch / "rdm1.npy"`

(the type cast to `Path` is fully redundant)

src/quemb/molbe/solver.py
```diff
@@ -908,35 +916,57 @@ def solve_block2(
...
+        rdm2_path = Path(scratch + "/rdm2.npy")
+        try:
```
mcocdawc, line 930: `assert` for this kind of check is dangerous.

it should be only used for checks which you always assume to be true and which are there to catch errors in developer assumptions.

As soon as you execute an optimized build `python -O` they are not executed!

I would do something akin to

```python3 
if not (rdm1_path.is_file() and rdm2_path.is_file()):
    raise FileNotFoundError("Error: `force_earlystop` requires "
                "manually specified fragment 1 and 2rdms!")
```

src/quemb/shared/helper.py
```diff
@@ -95,4 +95,7 @@ def ncore_(z: int) -> int:
 def delete_multiple_files(*args: Iterable[Path]) -> None:
```
mcocdawc, line 95: if it iterates now also into subdirectories, then this could be replaced by `shutil.rmtree`

tests/data/molecular_DMRG_test/f1/rdm1.npy
mcocdawc, line 1: are all these npy files necessary? or can they be removed?

Since you added the directory to the .gitignore I assume the latter.
Then you still have to remove and git commit the removal

src/quemb/molbe/mbe.py
```diff
@@ -172,9 +178,18 @@ def __init__(
...
-        else:
+            self.scratch_dir = WorkDir.from_environment(
+                cleanup_at_end=self.cleanup_at_end,
+            )
+        elif isinstance(scratch_dir, WorkDir):
             self.scratch_dir = scratch_dir
+
+        elif isinstance(scratch_dir, PathLike):
+            self.scratch_dir = WorkDir(
+                path=Path(scratch_dir),
```
mcocdawc, line 189: The cast to Path is redundant here. **but** this comment is anyway superfluos, since this elif branch should IMHO not exist.

mcocdawc (commented):
Thanks for making this test work!

See my comments below

I would await a comment of @mscho527 or @lweisburn regarding the adjusted energy

src/quemb/molbe/solver.py
mcocdawc: BTW: If you cast to be sure that you indeed get a specific type, I would add a type specifier, i.e.
```python3
rdm1_path: Path = ...
`` `

File: .github/workflows/quemb_unittest.yml [hunk 5]
```diff
@@ -79,10 +79,9 @@ jobs:
 
   testsuite:
     runs-on: ubuntu-latest
-    needs: analysis
     strategy:
       matrix:
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.12", "3.13"]
```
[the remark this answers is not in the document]
mscho527: Uh should we keep 3.10? 3.10 (and for a few months, even 3.9) has not reached EOL yet

tests/data/molecular_DMRG_test/f1/rdm1.npy
ShaunWeatherly: They are necessary for executing the DMRG unittests in the workflow without explicitly calling `block2`. As are the `dmrg.conf.target` files. The directory is added to `.gitignore` because executing the tests locally generates a plethora of other DMRG files that should not be included in the repo.

src/quemb/molbe/solver.py
ShaunWeatherly: Good point! I'll make the changes.

.github/workflows/quemb_unittest.yml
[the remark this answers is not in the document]
ShaunWeatherly: Ah this can be reverted, not a necessary change.

setup.py
```diff
@@ -22,6 +22,9 @@
         "numpy>=1.22.0",
         "scipy>=1.7.0",
         "pyscf>=2.0.0",
+        "intel-openmp==2021.4.0",
```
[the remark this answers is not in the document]
ShaunWeatherly: That would be correct, I forgot to remove these.

src/quemb/molbe/be_parallel.py
```diff
@@ -193,6 +201,28 @@ def run_solver(
         mch.mc1step()
         rdm1_tmp, rdm2s = mch.fcisolver.make_rdm12(0, nmo, nelec)
 
+    elif solver in ["block2", "DMRG", "DMRGCI", "DMRGSCF"]:
+        frag_scratch = WorkDir(
+            path=Path(scratch_dir / dname),
```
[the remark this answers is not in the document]
ShaunWeatherly: `cleanup_at_end` is not redundant, just consider the converse of your example: if `scratch_dir.cleanup_at_end` is `False`, one would also like the fragment scratch directories to have `scratch_dir.cleanup_at_end` to be `False`, whereas by default they would be set to True and the fragment scratch directories would be removed despite the user setting `cleanup_at_end` to False.

src/quemb/molbe/be_parallel.py
```diff
@@ -193,6 +201,28 @@ def run_solver(
...
+            cleanup_at_end=scratch_dir.cleanup_at_end,
+            allow_existing=scratch_dir.allow_existing,
+        )
+
+        assert isinstance(solver_args, DMRG_ArgsUser)
+        DMRG_args = _DMRG_Args.from_user_input(solver_args, mf_)
+
+        try:
```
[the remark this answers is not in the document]
ShaunWeatherly: Yup, it's been removed.

src/quemb/molbe/mbe.py
```diff
@@ -76,7 +77,8 @@ def __init__(
         restart_file: PathLike = "storebe.pk",
         nproc: int = 1,
         ompnum: int = 4,
-        scratch_dir: WorkDir | None = None,
+        scratch_dir: WorkDir | PathLike | None = None,
```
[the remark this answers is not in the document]
ShaunWeatherly: That's true, but I think the inclusion of `cleanup_at_end` at the `BE` class level is necessary. Otherwise, the user loses any control over cleanup unless they strictly provide `scratch_dir` as a `WorkDir`. This can be fixed just by allowing one to take precedence.

src/quemb/shared/helper.py
ShaunWeatherly: Perhaps it could, but I think you originally wrote this function so I'll leave removing it as a choice for you to make!

src/quemb/molbe/solver.py
```diff
@@ -355,7 +357,11 @@ def be_func(
             assert isinstance(solver_args, SHCI_ArgsUser)
             SHCI_args = _SHCI_Args.from_user_input(solver_args)
 
-            frag_scratch = WorkDir(scratch_dir / fobj.dname)
+            frag_scratch = WorkDir(
+                scratch_dir / fobj.dname,
+                cleanup_at_end=scratch_dir.cleanup_at_end,
```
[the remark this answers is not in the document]
ShaunWeatherly: See my comment above.

src/quemb/molbe/solver.py
```diff
@@ -401,7 +407,11 @@ def be_func(
             rdm1_tmp, rdm2s = ci.make_rdm12(0, nmo, nelec)
 
         elif solver in ["block2", "DMRG", "DMRGCI", "DMRGSCF"]:
-            frag_scratch = WorkDir(scratch_dir / fobj.dname)
+            frag_scratch = WorkDir(
+                scratch_dir / fobj.dname,
```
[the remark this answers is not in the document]
ShaunWeatherly: Again, see my comment above.

src/quemb/molbe/solver.py
```diff
@@ -893,7 +897,11 @@ def solve_block2(
     # pylint: disable-next=E0611
     from pyscf import dmrgscf  # noqa: PLC0415   # optional module
 
+    if ompnum is not None:
+        dmrgscf.settings.MPIPREFIX = "mpirun -np " + str(ompnum) + " --bind-to none"
+
     orbs = mf.mo_coeff
+    scratch = str(PurePath(frag_scratch))
```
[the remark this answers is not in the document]
ShaunWeatherly: You cannot actually, but maybe its just this circumstance. Doing so returns the error:
```
FileNotFoundError: [Errno 2] No such file or directory: "WorkDir(path=PosixPath('/home/shaunrw/quemb/tests/data/molecular_DMRG_test/f0'), cleanup_at_end=False, allow_existing=True)/dmrg.conf"
```

.github/workflows/quemb_unittest.yml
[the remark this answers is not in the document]
mscho527: Seems like this was not reverted yet?

.github/workflows/quemb_unittest.yml
```diff
@@ -92,34 +91,77 @@ jobs:
...
+    - name: Restore pip Cache
+      id: restore-pip-cache
+      uses: actions/cache/restore@v4
       with:
         path: ${{ env.pythonLocation }}
-        key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('dev-requirements.txt') }}
+        key: ${{ env.pythonLocation }}-${{ runner.OS  }}-pip-cache-${{ hashFiles('setup.py') }}}}
+
+    - name: Install pip
+      if: steps.restore-pip-cache.outputs.cache-hit != 'true'
```
mscho527, line 102: Shouldn't `pip` upgrade happen regardless of the cache hit?

.github/workflows/quemb_unittest.yml
```diff
@@ -92,34 +91,77 @@ jobs:
...
-        mv settings.py.example ${PYSCFHOME}/pyscf/dmrgscf/settings.py
         pip install .
 
+    - name: Save quemb Cache
+      if: steps.restore-quemb-cache.outputs.cache-hit != 'true'
+      id: save-quemb-cache
+      uses: actions/cache/save@v4
+      with: 
+        path: ${{ env.pythonLocation }}
+        key: ${{ steps.restore-quemb-cache.outputs.cache-primary-key }}
```
mscho527, line 132: Maybe I'm wrong, but isn't this potentially dangerous? I think we want to test installing `quemb` everytime, even if it adds a few seconds to the test suite. Since we are testing `quemb`, I don't feel comfortable caching it based on hashes that only look at specific files.

.github/workflows/quemb_unittest.yml
```diff
@@ -92,34 +91,77 @@ jobs:
...
+    
+    - name: Restore block2 Cache
+      id: restore-block2-cache
+      uses: actions/cache/restore@v4
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.pythonLocation }}-${{ runner.OS }}-block2-cache-${{ hashFiles('setup.py') }}-${{ hashFiles('dev-requirements.txt') }}-${{ hashFiles('test_requirements.txt')}}}}
```
mscho527, line 139: Does this key depend on the contents of `block2` repository? In other words, if something changes on `pyscf/dmrgscf` or pypi release of `block2`, does this key logic distinguish the change?

.github/workflows/quemb_unittest.yml
```diff
@@ -92,34 +91,77 @@ jobs:
...
+      if: steps.restore-block2-cache.outputs.cache-hit != 'true'
+      id: save-block2-cache
+      uses: actions/cache/save@v4
+      with: 
+        path: ${{ env.pythonLocation }}
+        key: ${{ steps.restore-block2-cache.outputs.cache-primary-key }}
 
     - name: Test with pytest
       run: |
+        pip install .
```
mscho527, line 160: Is this redundant? Line 124 installs from the repo as well.

example/molbe_h8_chemical_potential.py
```diff
@@ -33,7 +33,7 @@
 # Perform BE calculations with different fragment schemes:
 
 # Define BE1 fragments
-fobj = fragpart(be_type="be1", mol=mol)
+fobj = fragpart(be_type="be1", frag_type="autogen", mol=mol)
```
mscho527, line 36: `graphgen` can do BE1 as well, right? Do you want to add a block (like be2 and be3 below) that demonstrates `graphgen` for be1, too?

src/quemb/molbe/be_parallel.py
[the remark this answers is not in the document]
mscho527: Are you @ShaunWeatherly creating `WorkDir` objects for each fragment? Wouldn't we want one `WorkDir` object (i.e. `molbe.scratch_dir` and create subdirectories (`dname`) under here? Unless we want granular control over the cleanups (i.e. we want to clean up fragment 0 but leave fragment 1), I don't think multiple instances of `WorkDir` would be necessary?

src/quemb/molbe/fragment.py
```diff
@@ -137,6 +138,7 @@ def __init__(
             self.center_atom = fragment_map.center_atoms
             self.centerf_idx = fragment_map.centerf_idx
             self.ebe_weight = fragment_map.ebe_weights
+            self.dnames = fragment_map.dnames
```
mscho527, line 141: Is `dnames` just the order within the list of fragments (i.e. fragment 0, fragment 1, etc)? It is a bit unclear how where this is being used, since `dnames` seems to only exist in `graphgen`. (i.e. if `autogen` is called, this would be missing, right?)

src/quemb/molbe/mbe.py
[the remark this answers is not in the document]
mscho527: My 2ct: If the user wants to control the cleanup part, I think they should be forced to explicitly create `WorkDir` with `cleanup_at_end=False`.

src/quemb/molbe/solver.py
```diff
@@ -893,7 +897,11 @@ def solve_block2(
     # pylint: disable-next=E0611
     from pyscf import dmrgscf  # noqa: PLC0415   # optional module
 
+    if ompnum is not None:
+        dmrgscf.settings.MPIPREFIX = "mpirun -np " + str(ompnum) + " --bind-to none"
```
mscho527, line 901: I'm confused about this line. Are you using `ompnum` as the number of OpenMP threads or MPI processes?

src/quemb/shared/external/dmrgscf_settings.py
```diff
@@ -0,0 +1,17 @@
...
+# Configs will be used in dmrgsci.py and chemps2.py
+#
+
+import os
+
+# For Block and Block2 solvers
+BLOCKEXE = "block2main"
+BLOCKEXE_COMPRESS_NEVPT = "block2main"
+BLOCKSCRATCHDIR = str(os.getpid())
+BLOCKRUNTIMEDIR = str(os.getpid())
```
mscho527, line 12: I don't know how `block` is called, but shouldn't these be directories, not PIDs?

src/quemb/shared/helper.py
mscho527: I'm guessing that oskar's comment was about if else blocks added in 98-101?

src/quemb/shared/manage_scratch.py
```diff
@@ -84,12 +89,15 @@ class WorkDir:
 
     path: Final[Annotated[Path, "An absolute path"]] = field(converter=_get_abs_path)
     cleanup_at_end: Final[bool] = True
+    allow_existing: Final[bool] = False
```
mscho527, line 92: Is there a particular reason why the default value of `allow_existing` has to be `False`? Does this conflict with mixed-basis code and require modification there as is? @lweisburn

tests/data/molecular_DMRG_test/f0/dmrg.conf
```diff
@@ -0,0 +1,21 @@
...
+60    100  1.0000e-08  0.0000e+00
+end
+twodot_to_onedot 50
+orbitals FCIDUMP
+maxiter 60
+sweep_tol 1.0000e-07
+outputlevel 2
+hf_occ integral
+twopdm
+prefix  /home/shaunrw/quemb/tests/data/molecular_DMRG_test/f0
```
mscho527, line 18: Are these `*conf` and `*conf.target` files necessary? These contain path to your home directory.

tests/dmrg_molBE_test.py
```diff
@@ -19,60 +21,109 @@
...
+            pop_method="lowdin",
+            scratch_dir=scratch,
+        )
+        mybe.oneshot(
+            solver="block2",
+            solver_args=DMRG_ArgsUser(
+                maxM=100,
+                max_iter=60,
+                max_mem=3,
+                force_earlystop=force_earlystop,
```
mscho527, line 93: Is `force_earlystop` a keyword that you would use normally, or is this just necessary because we want to test the interface without invoking `block2`? If it is the latter, I'm not too sure what I feel about this -- I'll have to think a bit more.

tests/dmrg_molBE_test.py
```diff
@@ -19,60 +21,109 @@
...
+            "be1",
+            "dmrg_h8_pipek_interface",
+            "hchain_simple",
+            -4.20236532,
+        )
+
+    @unittest.skipIf(
+        os.getenv("GITHUB_ACTIONS") == "true",
+        "This test cannot be run in github actions.",
+    )
```
mscho527, line 47: Is there an easy way to detect whether `block2` is installed or not? If so, I think replacing `GITHUB_ACTIONS` check with the check would be nice.

mscho527 (changes requested):
I left some comments regarding some questions and confusions, but looks solid so far!

src/quemb/molbe/mbe.py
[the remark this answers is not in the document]
mcocdawc: > This can be fixed just by allowing one to take precedence.

Unfortunately not;

if `cleanup_at_end` has precedence, then
```python3
BE(..., WorkDir("/tmp", cleanup_at_end=False)) 
```
would unexpectedly cleanup the WorkDir

if `WorkDir` has precedence, then
```python3
BE(..., cleanup_at_end=False, WorkDir("/tmp")) 
```
would unexpectedly cleanup the WorkDir.

You would have to create a default `cleanup_at_end=None, WorkDir=None` and make distinctions based on if user input was given.
Then you would also need to thoroughly document the precedence.

And even if one argument has precedence, it still means that one can write code that is ambiguous to others, i.e.:
```python3
BE(..., cleanup_at_end=True, WorkDir("/tmp", cleanup_at_end=False)) 
```

I really do not see the problem of users being forced to pass a WorkDir object if it means that:
- it is not possible to write ambiguous code,
- there is no necessity to document precedence, and
- no involved checking for user-input is necessary.

src/quemb/molbe/solver.py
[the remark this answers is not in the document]
mcocdawc: of course :-D

src/quemb/shared/helper.py
mcocdawc: I'll have a look

src/quemb/molbe/solver.py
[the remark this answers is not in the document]
mcocdawc: oh my bad, I thought I had something implemented as:

```python3
    def __str__(self) -> str:
        return str(self.path)
```

but on second thought it is perhaps anyway better if it is not implemented. 

You can still just call `str(frag_scratch.path)`, then you don't have to cast twice.

.github/workflows/quemb_unittest.yml
ShaunWeatherly: This cache is really just for the dependencies, we still run `pip install .` at the beginning of the testsuite to install quemb itself. I can change the name if it's misleading.

.github/workflows/quemb_unittest.yml
ShaunWeatherly: Those changes would need to be reflected in the list of `test_requirements.txt` dependencies, but otherwise no explicit information from block2 is used for key generations. This is also true for all of the other dependencies, however.

.github/workflows/quemb_unittest.yml
ShaunWeatherly: If there are any changes to quemb itself, this is the line that updates those parts of the code. The previous `pip install .` above is really only used for installing/caching the dependencies in `setup.py`, hence why it's hash is used in the key.

tests/dmrg_molBE_test.py
ShaunWeatherly: The installation of block2 is not the problem, it's that the test simply _cannot_ be run in github actions. Oskar and I spent a lot of time trying to find a better work around, but to no avail. This seems to be the easiest solution.

tests/dmrg_molBE_test.py
ShaunWeatherly: I agree, its an ugly workaround. One could technically use it for any calculation where you want to skip calling `block2` but still proceed with the rest of the workflow, but that's not the most practical utility...

tests/data/molecular_DMRG_test/f0/dmrg.conf
ShaunWeatherly: They're used in the unittests for checkong that the `dmrg.conf` files generated by the wrapper are correct. These files are the interface to `block2`, so if someone breaks the wrapper this will be the first place it shows up. (The scratch directory given by `prefix` is excluded from the check, of course)

src/quemb/shared/manage_scratch.py
ShaunWeatherly: This is simply how Oskar has setup `WorkDir`, where the scratch directory provided by the user was _required_ to be empty (otherwise it would raise a ValueError). This gives you the option to turn off that requirement by setting `allow_existing=True`

.github/workflows/quemb_unittest.yml
mscho527: Yes. I think the name is misleading. If the purpose is to just install the dependencies, should line 124 (`pip install .`) be removed as well?

.github/workflows/quemb_unittest.yml
mscho527: Okay, this raises a good point then. @mcocdawc If we want to see intended results with the cache, I think we need to actually give specific version number in `test_requirements.txt` even if the packages are mentioned in `setup.py`. As it is currently set up now, even if new versions are uploaded on pypi (or github for `libdmet`), cache key would stay same (and thus we wouldn't be able to pull newer version on pypi without changing `setup.py`). I propose that we fix this in a separate PR.

Similarly, @ShaunWeatherly I think `dmrgscf` should also reference a specific tag or commit. `block2` should also have a version specified (both in the workflow and the key).

.github/workflows/quemb_unittest.yml
mscho527: I'm not sure why we would want to install and reinstall if there are any changes. The hash key in the quemb cache workflow above cannot detect changes made in the quemb code. I think we should install quemb from the latest commit once other dependencies are resolved (through cache, pip, or what not)

src/quemb/shared/manage_scratch.py
mscho527: @mcocdawc Was there a reason why you required scratch directory to be empty?

tests/dmrg_molBE_test.py
mscho527: Maybe Oskar and you discussed this already, but what do you mean by 'cannot'? I trust you two, but just curious.

Anyway, if I understood correctly, this test essentially tests i) all the way upto calling `block2` and ii) after `block2` finishes and `quemb` should do some post-processing, right? I worry that as is, the tests can make us overconfident. If at any point no one performs BE-DMRG calculations anymore, CI test on itself would no longer automatically test the full pipeline. I think this aspect of DMRG test routine should be noted in the docs.

.github/workflows/quemb_unittest.yml
```diff
@@ -79,10 +79,9 @@ jobs:
 
   testsuite:
     runs-on: ubuntu-latest
-    needs: analysis
```
mscho527, line 82: Reminder to restore `analysis` requirement before merge

src/quemb/molbe/be_parallel.py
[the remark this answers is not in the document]
mcocdawc: having a normal subdirectory would also not force you to keep track of `scratch_dir.cleanup_at_end`.
If `scratch_dir.cleanup_at_end` is True the subdirectory will be deleted anyway and if `scratch_dir.cleanup_at_end`  is False, then it will be preserved.

.github/workflows/quemb_unittest.yml
mcocdawc: May I understand what the point of this caching is?
Getting cache invalidation correct is really hard and the proposed setup will force us to regularly manually update the `requirements` files.
If performance of the test suite is the only reason I fear it might be a bit unnecessarily complicated.
I get why it makes sense for projects with larger dependencies or where they have to compile dependencies, but our build process from scratch is so fast, that I just would do it from scratch.

src/quemb/shared/manage_scratch.py
```diff
@@ -73,6 +73,11 @@ class WorkDir:
     then the cleanup is performed when leaving the ContextManager.
     Again, assuming that :python:`cleanup_at_end` is true.
 
+    If :python: `allow_existing` is False (default), then the directory
+    specified by :python: `path` must be empty.
+    NOTE: Use caution when setting this to True, as any files in the non-empty
+    directory may be overwritten!
```
mcocdawc, line 79: Even more dangerous is the fact that the directory gets cleaned up, if `cleanup_at_end=True`.
i.e. 
```python3
WorkDir("/home/${USER}")
```
will clean up the home directory of the user in the end.

-------------------------------------------------------------------------------------

I would use the sphinx note syntax https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

```rst
... note: Use caution ...
```

.github/workflows/quemb_unittest.yml
mscho527: Also, another question for @ShaunWeatherly -- if the test suite on Github Actions isn't invoking `block2`, do we need to install `block2`?

src/quemb/shared/manage_scratch.py
mcocdawc: Initially I required it to prevent silent overwriting of files, as @ShaunWeatherly pointed out in his note.
But even more dangerous is that the WorkDir is deleted in the end if `cleanup_at_end` is true which it is by default.
This means that `WorkDir("/home/${USER}")` would delete the home directory.

On the other hand @lweisburn and you (?) @mscho527 need an already populated WorkDir for mixed basis calculations..

So yes, there should be an `allow_existing` argument which should be `False` by default.
It should also be added to the `from_environment` constructor. 
I would then also expand the error message along the line of.
```python3
ValueError("scratch_area is not empty and allow_existing is False. "
              "You can set `allow_existing=True` in the constructor, but note that "
               "the working directory is deleted if `cleanup_at_end` is True, which "
               "it is by default."
    )
```

src/quemb/molbe/solver.py
mcocdawc: good catch

src/quemb/shared/manage_scratch.py
mscho527: Aha @mcocdawc I see! @lweisburn we should assess how to let this work with mixed basis code

lweisburn:
> Note to @lweisburn and @mscho527 , this PR corrects the definition of `BE.ebe_tot` to be the total BE energy. Previously, some parts of the code were using it as the correlation energy, including in the `misc.be2puffin` function. I've updated that function to explicitly use the correlation energy (which I assume was the original intent) and I just wanted to double check and make sure I didn't break anything!

Great, thanks for noting this! We'll make sure that this is consistent with other external (mixed-basis) code too

src/quemb/shared/manage_scratch.py
lweisburn: With the mixed basis code, we will have to use this option discussed above with `allow_existing` to be `True`, but it makes sense that for other uses for it to be `False` by default

.github/workflows/quemb_unittest.yml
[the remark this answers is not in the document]
ShaunWeatherly: Done.

src/quemb/molbe/mbe.py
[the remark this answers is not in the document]
ShaunWeatherly: Allowing the user to pass a string as a scratch directory is a _near-universal_ function of most computational chemistry codes. Talking with Minsik and Leah, I think we all agree that we should preserve this functionality. As is, the user can specify `scratch_dir` as _either_ a `str` or a `WorkDir`. 

With respect to `cleanup_at_end`, I think allowing one to take precedence is the most straightforward workaround while still preserving functionality.

src/quemb/molbe/be_parallel.py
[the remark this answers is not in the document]
ShaunWeatherly: Sure, one could just use `PathLike`s for the fragment scratch directories, but I think using `WorkDir` instead keeps the code more internally consistent. The most elegent workaround would be to add a `create_subdirectory` class method to `WorkDir`, which inherits the properties of the base class.

src/quemb/molbe/solver.py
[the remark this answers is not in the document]
ShaunWeatherly: Done.

.github/workflows/quemb_unittest.yml
ShaunWeatherly: Restored.

src/quemb/shared/manage_scratch.py
ShaunWeatherly: Done.

src/quemb/molbe/be_parallel.py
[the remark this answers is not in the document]
mscho527: @mcocdawc @ShaunWeatherly is this resolved? what did we conclude to do for this?

example/molbe_h8_chemical_potential.py
mscho527: Sure, don't you also want to show that BE1 can be done with graphgen in this example file for consistency, though? Unresolving this

mscho527 (changes requested):
Seems like there are few points remaining that we want to resolve before merging this. Thanks!

Issue #61, opened 2024-12-11
ewald summation for gamma point


mscho527:
`kbe.pbe.py:ewald_sum` looks like it was modified from the old code for the gamma point calculations. Investigate this and make necessary modifications.

mscho527:
This issue perhaps can be resolved after v0.1. Updating the milestone label

PR #97, opened 2025-01-22
Cheaper test variants for Github Actions

mscho527:
Try to reduce time for expensive tests
Use `exxdiv=None` before #61
Ask `pytest` to always log expensive tests and time breakdown

7min / tests --> 5.5 min on github actions; save cpu hrs, save the planet :)

### DIFF
--- a/.github/workflows/quemb_unittest.yml
+++ b/.github/workflows/quemb_unittest.yml
@@ -117,6 +117,7 @@ jobs:
       run: |
         cd tests
         QUEMB_SKIP_EXPENSIVE_TESTS=true pytest --durations=0 --durations-min=1.0 --doctest-modules --junitxml=junit/quemb-test-results_${{ matrix.python-version }}.xml
+        # If --duration=0, pytest reports all tests that took longer than what was given with durations-min=[time in seconds]
 
 
     - name: Upload pytest junit results
--- a/.github/workflows/quemb_unittest.yml
+++ b/.github/workflows/quemb_unittest.yml
@@ -116,7 +116,7 @@ jobs:
     - name: Test with pytest
       run: |
         cd tests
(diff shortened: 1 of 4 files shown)


.github/workflows/quemb_unittest.yml
```diff
@@ -116,7 +116,7 @@ jobs:
     - name: Test with pytest
       run: |
         cd tests
-        QUEMB_SKIP_EXPENSIVE_TESTS=true pytest --doctest-modules --junitxml=junit/quemb-test-results_${{ matrix.python-version }}.xml
+        QUEMB_SKIP_EXPENSIVE_TESTS=true pytest --durations=0 --durations-min=1.0 --doctest-modules --junitxml=junit/quemb-test-results_${{ matrix.python-version }}.xml
```
mcocdawc, line 119: isn't the minimum `--durations-min` redundant, if you don't report any durations?

mcocdawc (commented):
just a short question about the pytest call. Otherwise looks good to me

.github/workflows/quemb_unittest.yml
mscho527: Oh yeah I also find this counterintuitive lol pytest reports `n` longest tests normally. If `n` is zero, it reports all tests that took longer than what was given with `durations-min`

mscho527: And just an aside, `durations-min` takes values in seconds haha min probably stands for minimum, not minutes

mcocdawc: ok then the change makes sense. 
I would perhaps add a comment explaining that `--durations=0` is indeed what we want and not an oversight.

mcocdawc (approved):
cool cool nice nice

.github/workflows/quemb_unittest.yml
mscho527: Sure will do!

PR #98, opened 2025-01-22
Add QMMM/BE (QM/QM/MM) example

lweisburn:
Added a new QM/MM example code, using CCSD with STO-3G for the QM region of an octane molecule surrounded by random point charges


mscho527 (approved):
cool nice nice nice 🍪

PR #108, opened 2025-02-08
fixed types

mcocdawc:
- Use the AtomIdx from chemcoord, which is the one returned by Cartesian.get_bonds()
- Do a few fixes to make pandas happy

### DIFF
--- a/docs/source/nitpick-exceptions
+++ b/docs/source/nitpick-exceptions
@@ -3,3 +3,4 @@ py:class numpy.float64
 py:class OrderedSet
 py:class ordered_set.OrderedSet
 py:class Number
+py:class AtomIdx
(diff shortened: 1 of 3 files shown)


lweisburn (approved):
ignoring the `arg-type` when performing `elements.loc[:, "atomic_radius_cc"].update(in_vdW_radius)` to avoid mypy issues seems fine to me, if you're okay not enforcing some type there!

mcocdawc:
> ignoring the `arg-type` when performing `elements.loc[:, "atomic_radius_cc"].update(in_vdW_radius)` to avoid mypy issues seems fine to me, if you're okay not enforcing some type there!

Yeah, that was a deliberate decision. The type hinting of pandas is in a separate repo, namely `pandas-stubs`. They explicitly say themselves, that they occasionally accept a narrower set of types compared to what pandas actually supports.
We can freely accept floats, integers both builtins and numpy ones.


```python3
from numbers import Real
import numpy as np

isinstance(float(), Real) is True
isinstance(int(), Real) is True
isinstance(np.float64(), Real) is True
isinstance(np.int64(), Real) is True
```

PR #112, opened 2025-02-12
Updates and cleanup for `graphgen`.

ShaunWeatherly:
=====New Changes

****1.** New ASCII visualization for fragment subgraphs.** 

When `VERBOSE` and using `graphgen`, the fragment adjacency graphs will be printed in a simplified string format. For example, a hydrogen chain with BE2-type fragments will yield:
```
VERBOSE: Fragment Connectivity Graphs -------------------------------------------
(Center sites within a fragment are [bracketed])
Frag `f0`:
╙── [H0]
    └── [H1]
        └── H2
Frag `f1`:
╙── H1
    └── [H2]
        └── H3
...
```
Some notes:

- Atom labels are indexed by the order in which they appear in the `.xyz`. 
- Atoms corresponding to _center sites_ are bracketed.
- Labels are connected by edges if and only if they are contained within the same fragment. 
- This notation does not indicate the _origin_ atom.

****2.** `FragmentMap` has been removed and the outputs of all fragmentation codes are now standardized.** 

Also, the fragmentation modules have been reorganized to be a bit more intuitive: `mbe.fragment.py` contains `FragPart` and serves as a common interface to all the fragmentation codes, which are now fully disentangled and live in their own modules: `autofrag.py`, `graphfrag.py`, and `chemfrag.py`.

*NOTE: This change is analagous to how `kbe.fragment` was already structured - i.e. now both periodic and molecular `fragment` modules are organized consistently.

****3.** Fix shared centers in `graphgen` and enables the full testsuite in `fragmentation_test`.** 

Namely, `test_shared_centers_autocratic_matching` now passes for `graphgen`.

===== Old Changes

**1.** Addresses Oskar's latest concern by allowing `remove_nonnunique_frags` to return a single fragment (See BE6 for octane below as an example). Note that the single fragment case still returns an error in later parts of the workflow:

```
File "/home/shaunrw/miniconda3/envs/quemb/lib/python3.11/site-packages/quemb/molbe/pfrag.py", line 454, in schmidt_decomposition
    Denv = Dhf[Env_sites, Env_sites.T]
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: arrays used as indices must be of integer (or boolean) type
```

(ie there are no environment sites when one fragment encompasses the entire system) and this is the same as for the other fragmentation codes.

**2.** Now uses a dynamic default `cutoff`. After BE3, the cutoff used when computing shortest paths in the adjacency graph will nowscale with the _n_ in BE_n_. Also, `cutoff` is now in units of Angstrom (previously it was Angstrom**2). 

**3.** Adds functions for exporting and visualizing the fragment connectivity graph in `FragmentMap.export_graph()`. Some example adjacency graphs for octane (BE2-BE6) are provided in `quemb/example/figures`. You can generate your own using `fragpart.export_graph_to`:

```
fobj = fragpart(
    be_type="be3",
    frag_type="graphgen",
    mol=mol,
    export_graph_to=work_dir,
    )
```

The snippet above would yield the following for octane:
image

**4.** Also, `be_type` is now parsed more intelligently using regex, meaning you can specify something arbitrary like `BE(256)` and it is parsed correctly as the 256th coordination sphere.


ShaunWeatherly:
Note that I've also disabled `mypy` static analysis in the workflow - I recall us deciding that we weren't going to force static typing in this repo. let me know if something's changed, but I personally find it quite annoying and would vote for it being permanently removed from the workflow. This is something we can discuss more if needed.

mscho527:
> Note that I've also disabled `mypy` static analysis in the workflow - I recall us deciding that we weren't going to force static typing in this repo. let me know if something's changed, but I personally find it quite annoying and would vote for it being permanently removed from the workflow. This is something we can discuss more if needed.

Based on my recollection of our discussions, we want to keep the `mypy` static analysis to catch obvious errors. We should keep it in the workflow.

We agreed that we will not force `mypy` rules for new development, so that it does not hinder new contributions to the code base. As we've done for a number of previous PRs, this is done by explicitly marking the relevant files as exceptions.

ShaunWeatherly:
Alright what you've both just said seems to be in conflict so I'm confused. 

According to @mscho527 , we agreed that static typing ** will not** be enforced for new files, but according to @mcocdawc we agreed that static typing **will** be enforced for new files. 

Currently, we take an "opt-out" approach, and it seems _most_ of the files in QuEmb have been added to the blacklist in `mypy.ini`. This seems a bit silly, perhaps we can use an "opt-in" approach instead given that the majority of files are excluded from mypy checking?

mcocdawc:
> [...] but according to @mcocdawc we agreed that static typing will be enforced for new files. [...]

That is not what I said, on the contrary it is not enforced and you can easily switch it off. As I wrote:

> [...] But you can easily have a more tolerant check for new files, by modifying the mypy.ini. [...]

Correct me if I am wrong, but to my ears that does not sound like a rule that is enforced.

So the only question is: what should be the default for new files?
It boils down to what the majority of us will use. I personally always want to use type-hinting and it is a nice reminder if I forgot an annotation.
AFAIK @mscho527  does so as well and @lweisburn started to use it.

> [...] and it seems most of the files in QuEmb [...]

Basically all the old files have been added. Whenever I added type hints to old files I removed them from the list. IMHO we shall aim to fully type hint new code and gradually add type hints to old code. But again, if you don't want to use this feature you don't have to.

mscho527:
I support the "opt-out" approach, with an understanding that no questions will be asked if new files are added to the ignore list.

src/quemb/molbe/autofrag.py
```diff
@@ -181,7 +181,7 @@ def export_graph(
         plt.tight_layout()
         plt.legend(patches, self.dnames, loc="upper left")
         plt.axis("off")
-        plt.savefig(outdir + outname + ".png", dpi=1500)
+        plt.savefig(outdir + outname + ".png", dpi=1500) # type: ignore[operator]
```
mcocdawc, line 184: If you don't want to use type hints, then please just remove them from the declared interface of your function, but if they are present they should be correct.

Here the types are not correct and the ignored error is actually a proper bug at runtime.

`outdir` is declared as `PathLike` to the user and can for example be a `pathlib.Path`.
If you do:
```python3
pathlib.Path(".") + "testfile.png"
```
You will get an error 
```text
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

So I would do one of the following options (in that order):
1. accept only `pathlib.Path` in the interface and work with that, i.e. `outdir /  f"{outname}.png"`
2. cast to `pathlib.Path` and work with that, i.e. `outdir /  f"{outname}.png"`
3. accept only `str` in the interface and work with that, i.e. f"{outdir}/{outname}.png"

Currently you promise the calling side to accept `pathlib.Path` while you statically know that it will fail at runtime.

I did not check the other `type: ignore` directives, but they should only be used if you know that `mypy` flags a false-positive.

mcocdawc: which is btw a prime example of why the whole type-checking is a good idea.

example/figures/AdjGraph_be5.png
mscho527, line 1: Is there an option in `graphgen` to place the hydrogen atoms to their connected heavy atoms? This BE5 fragmentation does not exactly match what we reported in papers, right? (i.e. f0 does not contain H21, 23, & 24 even though C19 is part of the fragment)

src/quemb/molbe/autofrag.py
```diff
@@ -107,14 +116,82 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
 
+    def export_graph(
+        self,
+        outdir: PathLike,
+        outname: str = "AdjGraph",
+        cmap: str = "cubehelix",
+        node_position: str = "coordinates",
+    ) -> None:
+        outdir = Path(outdir)
+        F = 2
```
mscho527, line 141: what is `F`? can you use more descriptive names for the variables?

src/quemb/molbe/autofrag.py
```diff
@@ -107,14 +116,82 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
+                    map["coord"][1] + (map["coord"][2] / F),
+                )
+                for _, map in self.adx_map.items()
+            ]
+        elif node_position in ["spring"]:
+            pos = nx.spring_layout(G, seed=3068)
+
+        __, _ = plt.subplots()
+        options = {"edgecolors": "tab:gray", "node_size": 800, "alpha": 1}
+        arc_rads = np.arange(-0.3, 0.3, 0.6 / len(c), dtype=float)
```
mscho527, line 164: Perhaps expose these to the function args or kwargs?

src/quemb/molbe/autofrag.py
```diff
@@ -332,6 +417,12 @@ def graphgen(
     for adx, _ in enumerate(fragment_map.fs):
         fragment_map.dnames.append(str(frag_prefix) + str(adx))
 
+    if export_graph_to:
+        fragment_map.export_graph(
+            outdir=str(export_graph_to),
```
mscho527, line 422: Any reason to cast PathLike to str when the function accepts PathLike?

mscho527 (changes requested):
cool! some changes requested

mscho527:
one minor nitpicky comment -- when this PR is squash-merged, please use a descriptive name (than the name of this PR) for the merge commit

src/quemb/molbe/autofrag.py
```diff
@@ -1,8 +1,12 @@
 # Author: Oinam Romesh Meitei, Shaun Weatherly
 
+import re
 from copy import deepcopy
+from pathlib import Path
 from typing import Sequence
```
mcocdawc, line 6: Sequence should preferably imported from `collections.abc`.
https://docs.python.org/3/library/typing.html#typing.Sequence

src/quemb/molbe/autofrag.py
```diff
@@ -186,7 +264,12 @@ def graphgen(
     if iao_valence_basis is not None:
         raise NotImplementedError("IAOs not yet implemented for graphgen.")
 
-    fragment_type_order = int(be_type[-1])
+    fragment_type_order = int(re.findall(r"\d+", str(be_type))[0])
```
mcocdawc, line 267: wouldn't a simple `int(be_type[2:])` suffice?

src/quemb/molbe/autofrag.py
```diff
@@ -284,10 +367,12 @@ def graphgen(
                     fsites_temp = fsites_temp + deepcopy(list(fragment_map.sites[bdx]))
                     fs_temp.append(deepcopy(fragment_map.sites[bdx]))
                     fatoms_temp.append(bdx)
+                    edges_temp = edges_temp + list(nx.utils.pairwise(path))
```
mcocdawc, line 370: easier done via
`edges_temp.extend(nx.utils.pairwise)`

src/quemb/molbe/autofrag.py
```diff
@@ -107,14 +116,82 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
+                    map["coord"][0] + (map["coord"][2] / F),
+                    map["coord"][1] + (map["coord"][2] / F),
+                )
+                for _, map in self.adx_map.items()
+            ]
+        elif node_position in ["spring"]:
+            pos = nx.spring_layout(G, seed=3068)
+
+        __, _ = plt.subplots()
+        options = {"edgecolors": "tab:gray", "node_size": 800, "alpha": 1}
```
mcocdawc, line 163: If you call the plt.subplots like this and discard its return value, then I think the whole line can go.
Preferably, I would actually accept the return values
`fig, ax = plt.subplots()`
and then pass `ax` to the networkx plotting functions.

Then you could even return the fig and ax which would allow users to further customize the output.

Currently they would have to redo all steps, if they want to just change a label.

src/quemb/molbe/autofrag.py
```diff
@@ -332,6 +417,12 @@ def graphgen(
     for adx, _ in enumerate(fragment_map.fs):
         fragment_map.dnames.append(str(frag_prefix) + str(adx))
```
mcocdawc, line 418: `frag_prefix` is declared as string anyway. The cast is redundant.

src/quemb/molbe/autofrag.py
mcocdawc: To second @mscho527 . It is actually dangerous to cast to string here and I would just pass it on

src/quemb/molbe/autofrag.py
```diff
@@ -107,14 +116,82 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
+
+        G = self.adjacency_graph
+
+        if node_position in ["coordinates"]:
+            pos = [
+                (
+                    map["coord"][0] + (map["coord"][2] / F),
+                    map["coord"][1] + (map["coord"][2] / F),
+                )
+                for _, map in self.adx_map.items()
```
mcocdawc, line 157: for _, map in self.adx_map.items()

is equivalent to

`for map in self.adx_map.values()`

example/figures/AdjGraph_be5.png
ShaunWeatherly: There is currently no option to append hydrogen atoms to their connected heavy atoms, and `graphgen` will not strictly replicate `autogen` for anything larger than BE2 (although the results are very close, typically differing by an atom or two).

mcocdawc (changes requested):
Before diving into the details of the PR, I'd like to first address a major issue we also discussed in person.

I fully support the idea of separating graphgen and autogen into their own modules, this is a great improvement.
However, moving FragPart into fragment.py introduces a circular dependency, which is problematic.

A potential clean solution could be to move FragPart into its own dedicated
module. This would allow all fragmentation-related code to import it
consistently, maintaining modularity and avoiding circular imports.
It would also help preserve a self-documenting structure, because
the fragmentation algorithms indeed do depend on the structure of FragPart;
i.e. if the joint expected output changes, we have to modify all fragmentation algorithms.

Replacing the class with a dictionary removes the safety and self-documenting features of a class.
Instead of having a limited amount of clearly defined components, one has to
juggle all names in their heads.
Already just a bug from a typo such as "ebe_weight" -> "ebe_weights" would be unnecessarily easy to introduce.

Hence, please revert the move to a dictionary before we proceed further.
Let’s keep the clarity and safety of a structured class where it's really beneficial.

PR #116, opened 2025-02-19
Add rudimentary timings

mscho527:
Small PR to add rudimentary timings for the software paper

----------

I've used simple `time` module for this since we don't have a good logging infrastructure yet. Lmk if there is a better way to do this!


mscho527:
AFAIK, `sphinx` version cap in #100 should fix the docs test

src/quemb/molbe/mbe.py
```diff
@@ -230,6 +233,11 @@ def __init__(
             self.initialize(mf._eri, compute_hf)
         else:
             self.initialize(None, compute_hf, restart=True)
+        if settings.PRINT_LEVEL >= 10:
+            print(
+                f"Time to initialize BE object: {time() - init_start_time}",
+                flush=True,
```
mcocdawc, line 239: IMHO we overdo the flushing.

In this case I would just print. (also applies to the rest)

mcocdawc (commented):
Looks good.

It seems as if we will do this some more times. Perhaps something like the following class might be useful?

```python
@define
class Timer:
    """A simple timer class."""
    message: str = "Elapsed time"
    start: float = field(init=False, factory=time)

    def elapsed(self) -> float:
        return time() - self.start

    def str_elapsed(self) -> str:
        return f"{self.message}: {self.elapsed():.3f}"
```

which would be used as
```python
init_timer = Timer("Time to initialize BE object")
print(init_timer.str_elapsed())
```

mscho527:
> Looks good.
> 
> It seems as if we will do this some more times. Perhaps something like the following class might be useful?
> 
> ```python
> @define
> class Timer:
>     """A simple timer class."""
>     message: str = "Elapsed time"
>     start: float = field(init=False, factory=time)
> 
>     def elapsed(self) -> float:
>         return time() - self.start
> 
>     def str_elapsed(self) -> str:
>         return f"{self.message}: {self.elapsed():.3f}"
> ```
> 
> which would be used as
> 
> ```python
> init_timer = Timer("Time to initialize BE object")
> print(init_timer.str_elapsed())
> ```

Ah that's neat! I'll shamelessly copy this and add it to the PR

src/quemb/molbe/opt.py
```diff
@@ -174,24 +177,34 @@ def optimize(self, method, J0=None, trust_region=False):
...
                         "Error in density matching      :   {:>2.4e}".format(self.err),
                         flush=True,
                     )
+                    if settings.PRINT_LEVEL >= 10:
+                        print(iter_timer.str_elapsed())
                     print(flush=True)
                     if self.err < self.conv_tol:
                         print(flush=True)
                         print("CONVERGED", flush=True)
+                        if settings.PRINT_LEVEL >= 10:
```
mcocdawc, line 203: If you want to change the message later it is probably better to do:
```python
def str_elapsed(message: str | None = None) -> str:
    return f"{self.message if message is None else message}: {self.elapsed():.5f}"
```

and do
```python3
print(step0_timer.str_elapsed("Total time to complete BE optimization"))
```

I will approve anyway, so you can do whatever you prefer.

mcocdawc (approved):
If you want to change the message later it is probably better to do:
```python
def str_elapsed(message: str | None = None) -> str:
    return f"{self.message if message is None else message}: {self.elapsed():.5f}"
```
Then one can declare the components as Final and the class as 
`@define(frozen=True)`

and do
```python3
print(step0_timer.str_elapsed("Total time to complete BE optimization"))
```

I will approve anyway, so you can do whatever you prefer.

src/quemb/molbe/opt.py
mscho527: cool thanks! this is helpful. i'll reflect those changes and merge :)

PR #117, opened 2025-02-19
Sparse 2el integrals

mcocdawc:
Major:
Added support for jittable and sparse two-electron integrals.

Minor:
remove pylint from the CI pipeline.
With #113 being merged, the undefined names in imports can all be caught by mypy.
Furthermore `pylint` actually produces more false-positives than `pylint` if there is a delayed symbol definition shenanigan, as they do in `numba`.


src/quemb/molbe/sparse_2el_integral.py
mscho527, line 1: Thinking a bit ahead, would the 2e-integrals for the periodic system have a similar structure (but per k-point and complex elements)?

src/quemb/molbe/sparse_2el_integral.py
```diff
@@ -0,0 +1,93 @@
...
+
+    >>> assert g[1, 2, 3, 10] == 0
+    """  # noqa: E501
+
+    def __init__(self) -> None:
+        self._data = Dict.empty(*kv_ty)
+
+    def __getitem__(self, key: tuple[int, int, int, int]) -> types.float64:
+        idx = self.compound(*key)
+        return self._data.get(idx, 0.0)
```
mscho527, line 83: Would the boundary checking just require adding an exception, or am I missing anything? I don't have a good idea of whether using numba.typed.Dict instead of normal python dict may change something.

mscho527 (approved):
coool! minor questions, but i think the pr is good as is.

src/quemb/molbe/sparse_2el_integral.py
mcocdawc: More or less it would have the same structure, but you need a new hashing functions instead of the current `compound` one, since the k-point index does not permute with the orbital indices in the same way as they permute among each other.

src/quemb/molbe/sparse_2el_integral.py
mcocdawc: It has nothing to do with numba vs a normal dict.
First and foremost it would require the class to know about the dimensions in each direction.

For "normal" two electron integrals this code is easy to add.
It would only add a bit of overhead since you then have an additional conditional, but probably that does not matter too much performance-wise.

It gets really hairy if you go to 2el integrals of uneven size.
Let's say you have three AOs and one fragment MO
```text
g_{mu, nu, kappa, i}
```
then the last dimension can be much shorter. 
If you do the equivalent permutation
```text
g_{mu, nu, i, kappa}
```
then the second-to-last dimension is much shorter.

Getting these checks at runtime correctly would add too much overhead IMHO.

PR #118, opened 2025-02-24
added frozen_core support for chemfrag

mcocdawc:


tests/molbe_octane_get_rdms_test.py
ShaunWeatherly, line 1: Does it make more sense to keep this as `autogen` for now and instead add a new frozen core test to `test_chemfrag.py`?

ShaunWeatherly (changes requested):
Just one change and then this is good to go.

tests/molbe_octane_get_rdms_test.py
mcocdawc: Yeah, makes also sense. 

In the spirit of keeping the the integration tests together and to make it obvious that we expect them to return the same results, I wrote it simply as a loop:
```python3
    for frag_type in ["chemgen", "autogen"]:
        fobj = fragpart(be_type="be2", frag_type=frag_type, mol=mol, frozen_core=True)
```
Is that fine as well?

ShaunWeatherly: Sure that makes sense, just as long as long as `autogen` remains in the code we should continue to test it. This approach works as well. I think we should just be able to add `graphgen` to the loop, but I can make that change at a later date.

ShaunWeatherly (approved):
Yep, looks good!

tests/molbe_octane_get_rdms_test.py
mcocdawc: Great. :-) 

Re adding `graphgen` to the loop:
Don't you get different fragments with `graphgen` which then leads to slightly different energies?

PR #123, opened 2025-03-04
Screen orbital indices

mcocdawc:
The doc build fails because of a bug in sphinx https://github.com/tox-dev/sphinx-autodoc-typehints/issues/532


mcocdawc:
Do it cleanly in #128

Issue #121, opened 2025-02-27
optimization bugs pt. 2


lweisburn:
Problems trying to use `max_iter`: there is a small indexing issue if you try to set `max_iter` in `optimize` (or `max_space`, in `BEOPT`) that causes the code to crash. This should be a quick fix that actually allows for BE to run for a certain number of iterations.

lweisburn:
PR https://github.com/troyvvgroup/quemb/pull/124 fixes the issue!

Issue #120, opened 2025-02-27
optimization bugs pt. 1


lweisburn:
As currently written, we can input `J0` directly in the `optimize` function in `mbe`. This then is immediately overwritten by the `get_e_error_jacobian` function, which has the unexposed `jac_solver` keyword that really would be nice to have exposed. As such, we cannot use different starting jacobians in the optimization procedure, and we also can't actually give one explicitly when calling the optimization routine.

This should be fixed by simply exposing the `jac_solver` keyword, rather than `J0`, and doing some testing to make sure this is passed through the code correctly.

(There are also some issues with the docstrings for the optimization routines that should be fixed)

mscho527:
did #124 fix everything for this issue? @lweisburn

lweisburn:
> did [#124](https://github.com/troyvvgroup/quemb/pull/124) fix everything for this issue? [@lweisburn](https://github.com/lweisburn)

Yes, this PR #124 fixes the issue!

PR #124, opened 2025-03-07
small optimization bug fixes

lweisburn:
Make fixes for the optimization routine by:

1. exposing the QN jacobian input
2. fixing max_iter

(and potentially more)

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -23,4 +23,4 @@ ignore = [
         # if people import stuff with a leading underscore
         # they (should) know that this is private by convention
 ]
-preview = true
+preview = false
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -23,4 +23,4 @@ ignore = [
         # if people import stuff with a leading underscore
         # they (should) know that this is private by convention
 ]
-preview = false
+preview = false # ruff preview expects old python syntax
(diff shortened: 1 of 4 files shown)


lweisburn:
These are only small fixes to control allowed number of iterations and allow the jacobian in the optimization to be changed. More optimization work is necessary, but this should be helpful (for now) and address #120 and #121

.ruff.toml
```diff
@@ -23,4 +23,4 @@ ignore = [
         # if people import stuff with a leading underscore
         # they (should) know that this is private by convention
 ]
-preview = true
+preview = false
```
mscho527, line 26:
```suggestion
preview = false # ruff preview expects old python syntax
```

lweisburn: Done!

src/quemb/shared/external/optqn.py
```diff
@@ -396,6 +397,8 @@ def get_be_error_jacobian_selffrag(self, jac_solver="HF"):
         res_func = ccsdres_func
     elif jac_solver == "HF":
         res_func = hfres_func
+    else:
+        raise NotImplementedError("Jacobian solver option not implemented.")
```
mscho527, line 401: On an unrelated note, it looks like `get_be_error_jacobian_selffrag` isn't being used in the code. We may want to investigate this later.

mscho527 (approved):
cool cool cool!

letting this pr in before others in the queue, because this PR is necessary for the software paper

src/quemb/shared/external/optqn.py
lweisburn: Agreed! This should be part of the actual optimization code review and update

PR #126, opened 2025-03-21
fix error message to print correctly

lweisburn:
fix teeny bug from PR #124 🤦‍♀️

Now, the optimization code prints that the BE code doesn't converge after N iterations, when the convergence actually fails

### DIFF
--- a/src/quemb/molbe/opt.py
+++ b/src/quemb/molbe/opt.py
@@ -208,9 +208,10 @@ class BEOPT:
                             )
                         print(flush=True)
                         break
-                print(
-                    "BE DID NOT CONVERGE IN " + str(self.max_space) + " STEPS",
-                    flush=True,
-                )
+                if self.err >= self.conv_tol:
+                    print(
+                        "BE DID NOT CONVERGE IN " + str(self.max_space) + " STEPS",
+                        flush=True,
+                    )
         else:
             raise ValueError("This optimization method for BE is not supported")


mcocdawc:
I saw this PR too late in my mails, but wouldn't it perhaps make sense to raise an exception instead of mere printing?
Otherwise the user might easily miss the fact that the calculation actually failed.

One could even supply a custom exception such as 
```python3
class NotConvergedError(RuntimeError):
    pass
```

and do 
```python3
raise NotConvergedError
```

mscho527:
@mcocdawc oh i think that would be something we could discuss before making a decision. if we throw an exception, when someone has a script that does:
```
<do hf>
<do be1>
<do be2>
```
exception thrown on the be1 optimization, i think, would mean that the script exits without running be2. i tend to think that non-converged results are still numbers (attached with a less satisfactory be error, of course), so printing out error message felt apt for me.

mcocdawc:
Yes, some users prefer to proceed without fully converged numbers, while others do not. Ideally, both workflows should be supported—the only question is what the default behavior should be.

Personally, I would default to raising an exception while allowing an option like `continue_not_converged` to override this.

In the case that we do not want to throw an exception I would still not use `print`, but instead the `warnings` library of python. This has several advantages.
- The user can escalate the warning to an error if they wish to do so. Similar to exceptions one can also have different categories of (custom) warnings.
- It plays nicely with UNIX's STDERR redirection. I.e. it would be printed to a standard error file when using ` 2> calc.err`.
- It is printed a bit more expressively in the terminal, yellow colors and such.

mscho527:
Sure -- I think the default should be to not raise an exception because the numbers that are not fully converged are still BE energies with some signficiance. I do think that using warnings is a good idea! Shall we address that on a future PR?

mcocdawc:
Okidoki, I can take care of that in a new PR. (keep track of it in #127)

PR #129, opened 2025-03-21
Change default for type checking

mcocdawc:
- per the last discussion the type-annotation is now opt-in by default. (see the changes to the `mypy.ini`)

- Did a few automatic, cosmetic changes via `pyupgrade --py310-plus **/*py` 
    - more f-strings, set comprehension, using `dict` instead of `typing.Dict` for type annotations etc.


mypy.ini
```diff
@@ -1,33 +1,14 @@
...
-[mypy-tests.fragmentation_test,tests.chem_dm_kBE_test,tests.chempot_molBE_test,tests.dm_molBE_test,tests.dmrg_molBE_test,tests.eri_onthefly_test,tests.hf-in-hf_BE_test,tests.kbe_polyacetylene_test,tests.molbe_h8_test,tests.molbe_io_fcidump_test,tests.molbe_octane_get_rdms_test,tests.molbe_oneshot_rbe_hcore_test,tests.molbe_oneshot_rbe_qmmm-fromchk_test,tests.ube-oneshot_test,tests._expected_data_for_chemfrag,tests.test_chemfrag,tests.test_known_bug_autogen,tests.loc_test]
-
-    disallow_untyped_defs = False
-    check_untyped_defs = False
-
-[mypy-example.kbe_polyacetylene,example.molbe_dmrg_block2,example.molbe_h8_chemical_potential,example.molbe_h8_density_matching,example.molbe_hexene_oneshot_uccsd,example.molbe_io_fcidump,example.molbe_octane_get_rdms,example.molbe_octane,example.molbe_oneshot_rbe_hcore,example.molbe_oneshot_rbe_qmmm-fromchk,example.molbe_oneshot_ube_qmmm,example.molbe_ppp]
-    disallow_untyped_defs = False
-    check_untyped_defs = False
 
+# explicitly enable stricter tests for some files
```
lweisburn, line 8: This opt-in makes sense to me!

tests/molbe_octane_test.py
```diff
@@ -74,18 +74,18 @@ def test_cubegen() -> None:
     # Write cube file to a temporary location
     with tempfile.TemporaryDirectory() as tmpdir:
         write_cube(mybe, tmpdir, fragment_idx=[3], cubegen_kwargs=dict(resolution=5))
-        with open(os.path.join(tmpdir, "frag_3_orb_2.cube"), "r") as f:
+        with open(os.path.join(tmpdir, "frag_3_orb_2.cube")) as f:
```
lweisburn, line 77: to clarify! From what I understand, "r" is the default for `open`, so this is removed because it's the default?

mcocdawc: Exactly.

PR #130, opened 2025-03-25
removed chain from molbe

mcocdawc:
it is fully contained in both chemgen and graphgen, has known bugs and we would be forced to support it, if we keep it.


mscho527 (approved):
cool cool cool!

Issue #122, opened 2025-02-27
Wrong IAO indexing in autogen


mcocdawc:
@lweisburn @mscho527 @ShaunWeatherly 

In addition to the bug of the structural fragmentation #101 , there is a bug in autogen when performing an IAO calculation.

Again I am assuming the simple `octane.xyz` from our test suite.
We sort by atom first all `C`, then all `H`, and then by `y` coordinate which goes along the chain.

I assume the following differently prepared fragmentations:
```python
m = Cartesian.read_xyz("octane.xyz").sort_values(by=["atom", "y"])

frozen_core = False
large_basis = "cc-pvdz"
small_basis = "sto-3g"
n_BE = 2

autogen_frags = fragpart(
    mol=m.to_pyscf(basis=large_basis),
    frag_type="autogen",
    print_frags=False,
    be_type=f"be{n_BE}",
    frozen_core=frozen_core,
)
autogen_frags_small = fragpart(
    mol=m.to_pyscf(basis=small_basis),
    frag_type="autogen",
    print_frags=False,
    be_type=f"be{n_BE}",
    frozen_core=frozen_core,
)
autogen_IAO_frags = fragpart(
    mol=m.to_pyscf(basis=large_basis),
    frag_type="autogen",
    be_type=f"be{n_BE}",
    iao_valence_basis=small_basis,
    print_frags=False,
    frozen_core=frozen_core,
)

chem_frags_big = Fragmented.from_mole(
    mol=m.to_pyscf(basis=large_basis),
    n_BE=n_BE,
    frozen_core=frozen_core,
    iao_valence_basis=small_basis,
)
chem_frags_small = Fragmented.from_mole(
    mol=m.to_pyscf(basis=small_basis), n_BE=n_BE, frozen_core=frozen_core
)
```

The resulting indices are the following (ellipsis is to be contiguously continued)
```
autogen_frags.center_idx[0][0] == [0, 1, 2, ..., 23]
```
The meaning becomes clearer if looking at the new fragmented datastructure:
```
chem_frags_big.other_rel_AO_per_edge_per_frag
```
where the corresponding element is
```
{2: OrderedSet([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]),
  13: OrderedSet([14, 15, 16, 17, 18]),
  14: OrderedSet([19, 20, 21, 22, 23])}
```
i.e. atom 2 is carbon, atom 13 and 14 are hydrogens.

[...]

PR #131, opened 2025-03-26
Unified fragpart

mcocdawc:
Implemented a clear separation between `FragPart`, the class that holds the fragment data, and `fragmentate` (previously `fragpart`), the function that computes it.

Now `FragPart` documents its components and `fragpart` documents its arguments. (These two are not the same!)

This was done for both `kbe` and `molbe`.

This more explicitly shows which components the three fragmentation codes `autogen`, `graphgen`, and `chemgen` (have) to support. Relevant for #112 

Renamed all components that come out of  `autogen`, `graphgen`, and `chemgen`  in a consistent way.

------------------------------------------------

Additionally I removed `polychain` from the `kbe` `fragmentate`. It  was buggy and unused.


mscho527:
Before I review the PR in more detail, would it make sense to think whether we can come up with a good name for the class that holds the fragment data (i.e. `FragPart`)? I think we could use this opportunity to find a name that does not simply differ by upper/lower-case with `fragpart`.

src/quemb/molbe/autofrag.py
```diff
@@ -143,11 +258,33 @@ def remove_nonnunique_frags(self, natm: int) -> None:
...
+            ebe_weight=self.ebe_weight,  # type: ignore[arg-type]
+            Frag_atom=self.Frag_atom,  # type: ignore[arg-type]
+            center_atom=self.center_atom,  # type: ignore[arg-type]
+            hlist_atom=MISSING,
+            add_center_atom=MISSING,
+            frozen_core=frozen_core,
+            iao_valence_basis=None,
+            iao_valence_only=False,
+        )
+
```
mscho527, line 288: For the `MISSING` items, is the plan to add these in a later PR or w/ #112?

src/quemb/molbe/autofrag.py
```diff
@@ -92,14 +207,14 @@ class FragmentMap:
...
     center: list[Sequence[int]]
     centerf_idx: list[Sequence[int]]
-    ebe_weights: list[Sequence]
+    ebe_weight: list[Sequence]
     sites: list[Sequence]
     dnames: list[str]
-    fragment_atoms: list[Sequence[int]]
-    center_atoms: list[Sequence[int]]
+    Frag_atom: list[Sequence[int]]
+    center_atom: list[Sequence[int]]
```
mscho527, line 217: For the renamed attributes, could you also modify the docstrings? I am not sure how I feel about the name changes themselves -- @ShaunWeatherly any thoughts?

mscho527 (changes requested):
I have a few minor questions, but looks good overall!

src/quemb/molbe/autofrag.py
mcocdawc: Note that these are very similar to the `autogen` components and now just unified names.
Before the assignments read for example as:
```python3
self.ebe_weight = fragment_map.ebe_weights
```
where it is easy to miss the additional "s" on the right hand side.

src/quemb/molbe/chemfrag.py
```diff
@@ -964,7 +941,7 @@ def __len__(self) -> int:
         """The number of fragments."""
         return len(self.AO_per_frag)
 
-    def _match_autogen_output_no_iao(self) -> AutogenOutput:
+    def _match_autogen_output_no_iao(self) -> FragPart:
```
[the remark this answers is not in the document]
mcocdawc: I am absolutely in favour of removing it, if we are sure that the difference in IAO indexing in autogen is a bug. (#122 )
In this case I would also raise an exception when someone tries to use autogen and IAO indexing.

I had this first for debugging purposes to make sure that I can replicate previous behaviour.

src/quemb/molbe/autofrag.py
mcocdawc: I forward this question to @ShaunWeatherly .

Note that the components are not missing because of this PR, but that the refactoring just makes it explicit.

src/quemb/molbe/chemfrag.py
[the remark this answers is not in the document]
mscho527: I see! re: 2), I agree that we use some better name for the function. re: 3), I think we could ask @lweisburn?

src/quemb/molbe/autofrag.py
mscho527: I see! Could you reflect the changes in lines 167-20 (docstrings) as well?

mscho527 (approved):

src/quemb/molbe/autofrag.py
mcocdawc: the docstrings are also fixed now

src/quemb/molbe/chemfrag.py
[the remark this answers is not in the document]
mcocdawc: @lweisburn  wants to keep `wrong_iao_indexing` it for the moment to  be able to test old behaviour of the code.
hence I keep it for the moment, but keep track in #135 

I changed the names of the functions as per @mscho527 comment.

mcocdawc:
> Before I review the PR in more detail, would it make sense to think whether we can come up with a good name for the class that holds the fragment data (i.e. `FragPart`)? I think we could use this opportunity to find a name that does not simply differ by upper/lower-case with `fragpart`.

Let's wait for #112 to do everything in one go.

src/quemb/molbe/chemfrag.py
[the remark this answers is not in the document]
lweisburn: thanks, I appreciate this! I would like to have this for future debugging and to see from where differences in behavior arise in versions of the code

PR #133, opened 2025-03-31
ensure autocratic matching for chemgen

### LINKED ISSUE #132: chemgen be(3) bug
For some systems, there appears to be a double-counting bug when using BE(3) and `chemgen`. Calculations begin with a massive HF-in-HF error, SCF fails for all of the fragments, and the calculations with and without density matching fail entirely.

Example polypropylene system: (this also pops up for polyglycine)
```
46

C          0.00000	 -0.34300        0.00000
C          1.18819        0.34300        0.00000
C          2.37637	 -0.34300        0.00000
C          3.56456        0.34300        0.00000
C          4.75275	 -0.34300        0.00000
C          5.94093        0.34300        0.00000
C          7.12912	 -0.34300        0.00000
C          8.31731        0.34300        0.00000
C          9.50549	 -0.34300        0.00000
C         10.69368        0.34300        0.00000
C         11.88187	 -0.34300        0.00000
C         13.07005        0.34300        0.00000
C         14.25824	 -0.34300        0.00000
C         15.44643        0.34300        0.00000
C         16.63462	 -0.34300        0.00000
C         17.82280        0.34300        0.00000
H         -0.95003        0.20550        0.00000
H          0.00000	 -1.44000        0.00000
H          1.18819        1.44000        0.00000
C          2.37637	 -1.44000        0.00000
H          3.56456        1.44000        0.00000
H          4.75275	 -1.44000        0.00000
H          5.94093        1.44000        0.00000
C          7.12912	 -1.44000        0.00000
H          8.31731        1.44000        0.00000
H          9.50549	 -1.44000        0.00000
H         10.69368        1.44000        0.00000
C         11.88187	 -1.44000        0.00000
H         13.07005        1.44000        0.00000
H         14.25824	 -1.44000        0.00000
H         15.44643        1.44000        0.00000
C         16.63462	 -1.44000        0.00000
H         17.82280        1.44000        0.00000
H         18.77283	 -0.20550        0.00000
H         15.62581	 -1.79666        0.00000
H         17.13902	 -1.79667        0.87365
H         17.13902	 -1.79667	-0.87365
H         10.87306	 -1.79666        0.00000
H         12.38627	 -1.79667        0.87365
H         12.38627	 -1.79667	-0.87365
H          6.12031	 -1.79666        0.00000
H          7.63352	 -1.79667        0.87365
H          7.63352	 -1.79667	-0.87365
H          1.36757	 -1.79666        0.00000
H          2.88078	 -1.79667        0.87365
H          2.88078	 -1.79667	-0.87365
```

I didn't do a ton of digging, but the issue appears to be double-counting of some of the atoms. 

I think a few questions arise when I compare similar atoms. C2 and C20, for instance, are never origins in this system (which they shouldn't be). C2 is considered an edge in fragment 2 and center in fragment 1, while C20 is a center in both, even though they are in identical (opposite) positions compared to the C3 and C4 origins. When I look at the `ebe_weight` which come out of `chemgen` vs. `autogen`, the `chemgen` fragments of those fragments have more indices. The hf-in-hf error as a result is 100s of Ha off.

This makes a difference in the simple energy calculation as well as density matching (do we enforce matching for either of these atoms? To which fragment?). We have to decide whether to take all of the energy for a given atom for one fragment (rank the "more-important" fragment for each atom, so perhaps we always have one as a center and the other as an edge), or add in flexibility to `ebe_weigh`t to make sure we don't double count.

I believe `autogen` does that first option and just assumes the fragment in which an atom is closer to the origin is the more important fragment, but I haven't dug in too much. I am guessing this is much faster for @mcocdawc to think through and fix than me!

mcocdawc:
- [x] as discussed with @lweisburn in person, ensure that `add_center_atom` is as defined for the mixed-basis code

Mayor changes:

Enable and ensure autocratic matching for chemgen.
Previously, the case of a motif appearing as center in multiple fragments was not properly handled.
This fixes #132.

Democratic matching is not supported, but if it becomes necessary will be easy to add.

Minor changes:
- Better error message upon large HF in HF error

- The bug reported in #132 is also present in graphgen and also leads to a huge HF in HF error. I added a known to fail test for graphgen. Relevant for @ShaunWeatherly and #112


tests/fragmentation_test.py
```diff
@@ -1209,6 +1210,25 @@ def test_graphgen_autogen_octane_be2(self):
...
+
+        fobj = fragpart(mol, be_type="be2", frag_type="graphgen", print_frags=False)
+        mybe = BE(mf, fobj)
+
+        assert np.isclose(mf.e_tot, mybe.ebe_hf)
+
+        fobj = fragpart(mol, be_type="be3", frag_type="graphgen", print_frags=False)
+        mybe = BE(mf, fobj)
+
+        assert np.isclose(mf.e_tot, mybe.ebe_hf)
```
mcocdawc, line 1230: @ShaunWeatherly here graphgen will fail with a huge HF in HF error

src/quemb/kbe/pbe.py
```diff
@@ -637,11 +638,13 @@ def initialize(self, compute_hf: bool, restart: bool = False) -> None:
             self.ebe_hf = E_hf + self.enuc + self.E_core - self.ek
             print(
                 "HF-in-HF error                 :  {:>.4e} Ha".format(hf_err),
-                flush=True,
             )
-
+            if abs(hf_err) > 1.0e-1:
+                raise RuntimeError(
+                    "Too large HF-in-HF error. It makes no sense to continue."
```
lweisburn, line 644: I really like this message, and it is often true! We do have such a large error for mixed-basis BE stuff however, so I would downgrade this from RuntimeError to warn. (We could also remove it, though I am happy to instead downgrade it to emphasize that > 0.1 Ha is truly an awful error for the majority of calculations. Up to your discresion!)

src/quemb/molbe/chemfrag.py
```diff
@@ -675,6 +690,133 @@ def to_comma_output(seq: Sequence) -> str:
...
+    ) -> FragmentIdx:
+        """Assuming that :python:``center`` is shared across :python:``fragments``,
+        return the index of the fragment which represents it best.
+
+        This is done by finding the shortest path to each fragment's origin
+        and taking the fragment with the closest origin.
+        """
+        nx_graph: Graph = Graph(self.conn_data.bonds_motifs)  # type: ignore[arg-type]
+
+        def distance_to_fragment(i_frag: FragmentIdx) -> int:
```
lweisburn, line 721: Clarifying question: Is this `shortest_path_length` from networkx by number of intermediate nodes between the two end nodes? If this is the case, do we ever have a case in which an atom is graph-wise equally connected to two different origins, but one is closer in Cartesian space? 

Similarly, a case in which an atom is spatially closer to one of the origins, but with more intermediate nodes in that path? 

This may never matter or pop up in real molecules, or we may want to completely omit geometric info for the sake of the black-box definition, but I want to make sure I understand this correctly!

src/quemb/molbe/chemfrag.py
```diff
@@ -675,6 +690,133 @@ def to_comma_output(seq: Sequence) -> str:
...
+        If there is a motif, which appears as center in multiple fragments,
+        we will choose a fragment whose origin is closest to the center.
+        In this fragment it will stay a center, while the same motif
+        will become an edge in the other fragments.
+
+        For example, if we have the following nested structure
+
+        .. code-block:: text
+
+            1 - 2 - 3 - 4 - 5
```
lweisburn, line 742: I like this example, and I think it is sufficient, so please don't feel compelled to change this, especially as you clarify that this is a nested substructure.

If this were the full structure, this would be one BE(3) fragment, with no fragments about atoms 2 or 4. If you add two atoms, we would have the other two fragments described:

1 - 2 - 3 - 4 - 5 - 6 - 7
(               )|
(               )8
(               )|
(               )9
                  
                  
In which we are discussing atom 8 as a part of fragments centered on origins 3, 4, and 5

src/quemb/molbe/autofrag.py
```diff
@@ -889,6 +889,7 @@ class ChemGenArgs:
     treat_H_different: Final[bool] = True
     bonds_atoms: Mapping[int, set[int]] | None = None
     vdW_radius: InVdWRadius | None = None
+    autocratic_matching: bool = True
 
     #: This argument is not meant to be used by the user.
     #: If it is true, then chemgen adheres to the old **wrong** indexing
```
lweisburn, line 895: Does `autocratic_matching=True` means old, wrong indexing? I thought turning this own makes removes the option for shared centers?

src/quemb/molbe/mbe.py
```diff
@@ -886,12 +887,13 @@ def initialize(self, eri_, compute_hf, restart=False) -> None:
             hf_err = self.hf_etot - self.ebe_hf
             print(
                 "HF-in-HF error                 :  {:>.4e} Ha".format(hf_err),
-                flush=True,
             )
+            if abs(hf_err) > 1.0e-1:
+                raise RuntimeError(
+                    "Too large HF-in-HF error. It makes no sense to continue."
```
lweisburn, line 893: same as in the pbe version!

src/quemb/molbe/ube.py
```diff
@@ -373,17 +374,17 @@ def initialize(self, eri_, compute_hf):
             hf_err = self.hf_etot - (E_hf + self.enuc + self.E_core)
 
             self.ebe_hf = E_hf + self.enuc + self.E_core - self.ek
-            print(
-                "HF-in-HF error                 :  {:>.4e} Ha".format(hf_err),
-                flush=True,
-            )
+            print("HF-in-HF error                 :  {:>.4e} Ha".format(hf_err))
```
lweisburn, line 377: same again!

tests/test_chemfrag.py
```diff
@@ -210,3 +214,50 @@ def test_conn_data_manipulation_of_vdW():
     for i_carbon in m.loc[m.atom == "C"].index:
         # check if carbons are connected to everything
         assert {i_carbon} | conn_data.bonds_atoms[i_carbon] == set(m.index)
+
+
+def test_molecule_with_autocratic_matching():
```
lweisburn, line 219: Great tests! One element we may want to consider in future work for all of the fragmentation methods (including autogen) is the density matching conditions. BE(3) with matching is very expensive as a regular test, but we likely want to convince ourselves that each method can correctly do BE(32) at some point

In the current code, autogen fails matching with shape errors and as such is definitely wrong for BE(32), but I think some investigation of the chemgen results to make sure we're confident in them to use as a reference would be good future work!

lweisburn (changes requested):
Great fix to address atoms which are centers of multiple fragments!

src/quemb/molbe/autofrag.py
lweisburn: clarified (oops!)

src/quemb/molbe/chemfrag.py
mcocdawc: I also return the index `i_frag` now, to achieve unique ordering, even in this case.

(The code would have worked also without it)

The new interface is
```python3
        def distance_to_fragment(i_frag: FragmentIdx) -> tuple[int, int]:
            """Return the distance to the fragment with index ``i_frag``,
            as measured by the edge-number of the shortest path to the closest origin.
            Additionally return the index itsef to achieve unique ordering, if the
            distance is degenerate."""
```

src/quemb/kbe/pbe.py
mcocdawc: > We do have such a large error for mixed-basis BE stuff however,

Ok, then I change it to a warning in all cases.

src/quemb/molbe/chemfrag.py
mcocdawc: I keep the smaller block but changed it to
```
            --- 1 - 2 - 3 - 4 - 5 ---
                        |
                        6
                        |
                        7
```
and emphasize that it is part of a larger molecule

tests/test_chemfrag.py
mcocdawc: with the `QUEMB_SKIP_EXPENSIVE_TESTS` environment variable we can at least define expensive tests and just not execute them as part of the normal test-suite. 

(You know this :-), I am writing this more for reference.)

lweisburn (approved):
Looks great!

PR #141, opened 2025-04-08
Interface chemgen for periodic be1

### LINKED ISSUE #140: `autogen` with periodic systems return be2 fragments when be1 is requested
Quick temporary fix (until chemgen fully supports periodic systems) would be to use molecular chemgen results for be1, since be1 fragmentation should be identical for either open or periodic boundaries

mscho527:
Temporary fix for #140

closes #140

### DIFF
--- a/src/quemb/kbe/fragment.py
+++ b/src/quemb/kbe/fragment.py
@@ -1,10 +1,13 @@
 # Author(s): Oinam Romesh Meitei
 
 
+from warnings import warn
+
 from attrs import define, field
 from pyscf.pbc.gto.cell import Cell
 
 from quemb.kbe.autofrag import autogen
+from quemb.molbe.chemfrag import ChemGenArgs, chemgen
 from quemb.molbe.helper import get_core
 
 
@@ -74,6 +77,7 @@ def fragmentate(
     frozen_core=False,
     self_match=False,
     allcen=True,
+    print_frags=True,
 ):
     """Fragment/partitioning definition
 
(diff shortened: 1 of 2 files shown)


src/quemb/kbe/fragment.py
```diff
@@ -74,6 +77,7 @@ def fragmentate(
     frozen_core=False,
     self_match=False,
     allcen=True,
+    print_frags=True,
```
mcocdawc, line 80: I would either forward the `print_frags` argument also to `autogen` or not expose it.

src/quemb/kbe/fragment.py
```diff
@@ -167,6 +171,40 @@ def fragmentate(
             iao_valence_basis=iao_valence_basis,
             kpt=kpt,
         )
-
+    elif frag_type == "chemgen":
+        warn("Periodic BE1 with chemgen is a temporary solution.")
+        if kpt is None:
```
mcocdawc, line 176: I would put all the checks for invariants as early as possible.
This would also fix the surprising output when having `be_type != "be1"`.
Currently it would first warn
"Periodic BE1 with chemgen is a temporary solution."
and then 
raise the error that only `be_type = 'be1'` is supported.

mcocdawc (approved):
Looks overall good, as always :-)

I am pleasantly surprised that chemgen apparently just works in the periodic case as well.

In addition to my two specific comments below I would also add a small test to `chem_dm_kBE_test.py`

PR #144, opened 2025-04-09
Feat: chemfrag for periodic systems

mscho527:
Allow basic fragmentation functionalities of periodic systems using chemfrag; Advanced functionalities (i.e. 2D layers) are beyond scope of this PR

Permanent fix for #140


src/quemb/molbe/chemfrag.py
```diff
@@ -336,6 +338,98 @@ def from_mole(
...
+        if bonds_atoms is None:
+            # Add periodic copies to a fake mol object
+            # [a1, a2, a3] = ([0,0,0], [0,0,1], [0,1,0], [0,1,1],
+            #                 [1,0,0], [1,0,1], [1,1,0], [1,1,1])
+            # Eight copies of the original cell to account for periodicity
+            mol = Mole()
+            mol.unit = "Bohr"
+            lattice_vectors = cell.a if is_au(cell.unit) else cell.a * param.BOHR
+            offsets = np.array(
+                [
```
mcocdawc, line 390: Could it be, that the code even works out of the box for 2D layers?

If one rewrites the offset calculation as such
```python3
from itertools import combination
[
    sum(combination, start=np.zeros(3))
    for combination in chain(
        *(combinations(lattice_vectors, n) for n in range(len(lattice_vectors) + 1))
    )
]
```
then it automatically generalizes to both 2D and 3D

```
lattice_vectors = np.eye(3)
# gives
[array([0., 0., 0.]),
 array([1., 0., 0.]),
 array([0., 1., 0.]),
 array([0., 0., 1.]),
 array([1., 1., 0.]),
 array([1., 0., 1.]),
 array([0., 1., 1.]),
 array([1., 1., 1.])]

# while 
lattice_vectors = np.eye(3)[:2]
# gives
[array([0., 0., 0.]),
 array([1., 0., 0.]),
 array([0., 1., 0.]),
 array([1., 1., 0.])]
```

I think the rest should just work

mscho527: The code, as is, (hopefully) should work for both 2D and 3D -- we typically set one of the lattice vector to a crazy large vector to make pseudo-2D periodic systems for 2D systems. i.e., for 2D surface, the lattice_vector would be something like `[[2.3, 0., 0.,],[-1., 1.2, 0.,],[0.,0.,20.]]`

What I don't expect to support from this PR are 2D double layers and 2D double layers with tilt angles (both of which were introduced in earlier paper and implemented in autogen)

src/quemb/molbe/chemfrag.py
```diff
@@ -630,17 +727,30 @@ def from_mole(
...
+                    treat_H_different=treat_H_different,
+                    bonds_atoms=bonds_atoms,
+                    vdW_radius=vdW_radius,
+                ),
+                n_BE,
+                swallow_replace=swallow_replace,
+            )
+        elif type(mol) is Cell:
+            fragments = cls.from_conn_data(
+                mol.to_mol(),
```
mcocdawc, line 744: It would be nice to keep track of `mol` as cell instance.
Otherwise we lose information about lattice vectors etc.

This would require changing the type of `self.mol` to `Mole | Cell` ( as you already do in other occasions)

src/quemb/molbe/chemfrag.py
mcocdawc, line 1: Could you change the docstring at the top of this file to document that it also supports periodic systems.

src/quemb/kbe/fragment.py
```diff
@@ -237,34 +234,36 @@ def fragmentate(
...
-            args=ChemGenArgs(),
+            args=additional_args,
             iao_valence_basis=iao_valence_basis,
         )
-        molecular_FragPart = fragments.get_FragPart()
         if print_frags:
             print(fragments.frag_structure.get_string())
+        # Once periodic FragPart API is fixed,
+        # add _get_FragPart_no_iao equivalent in quemb.molbe.chemgen
+        mol_fragments = fragments.get_FragPart(
```
mcocdawc, line 252: It might be worth a thought to not allow the `_wrong_iao_indexing` in the periodic case. In the molecular part we only have it for testing purposes of previous code, but AFAIK there is no previous working periodic, IAO code.

mcocdawc (commented):
Very nice addition! Thank you very much.
 
I have a few comments sprinkled through the code and two architecture comments.

1. Since `chemfrag` is now neither purely molecular, nor purely periodic, I would move it to `shared`.

2. It might be a bit overkill, but it would actually be nice to not only declare the `mol` component as union type, but to have a Generic class with either `Mole` or `Cell` as upper bound. This way you encode, that in the periodic case all constructors only accept `Cell` and otherwise only `Mole`. In addition, down-stream code would be able to know if the `.mol` component is either of type `Mole` or of type `Cell` instead of the  union type.
The way to do it is roughly:
```python3

_T = TypeVar('_T', Mole, Cell)


class PurelyStructureFragmented(Generic[_T]):
    mol: Final[_T] = ....
    
    @classmethod
    def from_conn_data(
        cls,
        mol: _T, 
```

src/quemb/kbe/fragment.py
mscho527: How about we revisit this once we have a clearer picture of #125? Not allowing `_wrong_iao_indexing` should be rather simple at that point

mscho527:
> Very nice addition! Thank you very much.
> 
> I have a few comments sprinkled through the code and two architecture comments.
> 
> 1. Since `chemfrag` is now neither purely molecular, nor purely periodic, I would move it to `shared`.
> 2. It might be a bit overkill, but it would actually be nice to not only declare the `mol` component as union type, but to have a Generic class with either `Mole` or `Cell` as upper bound. This way you encode, that in the periodic case all constructors only accept `Cell` and otherwise only `Mole`. In addition, down-stream code would be able to know if the `.mol` component is either of type `Mole` or of type `Cell` instead of the  union type.
>    The way to do it is roughly:
> 
> ```python
> _T = TypeVar('_T', Mole, Cell)
> 
> 
> class PurelyStructureFragmented(Generic[_T]):
>     mol: Final[_T] = ....
>     
>     @classmethod
>     def from_conn_data(
>         cls,
>         mol: _T, 
> ```

Re 1: How about we do the move in a separate PR? As we discussed earlier, the move from `molbe` to `shared` will have nearly no impact on the public API. I'd prefer to keep the PR concise.
Re 2: Sure, I'll make those changes when I get a moment :)

src/quemb/molbe/chemfrag.py
```diff
@@ -325,6 +327,101 @@ def from_mole(
...
+            )
+            # Choose unique pairs
+            bonds_atoms = {}
+            for idx, bond in connectivity.bonds_atoms.items():
+                bonds_atoms.update(
+                    {
+                        idx % cell.natm: {bonded % cell.natm for bonded in bond}
+                        | bonds_atoms.get(idx % cell.natm, set())
+                    }
+                )
```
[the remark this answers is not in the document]
mscho527: Please correct me if I'm wrong, but wouldn't the simplified one not consider bonds starting from the periodic copies? i.e., when `i > len(cell.natm)`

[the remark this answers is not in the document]
mcocdawc: Perhaps I am now misunderstanding something, but any bond starting from the periodic copy and connecting to the original unit cell is also a bond starting from the original unit cell and connecting to the periodic copy.

To clarify one convection in chemcoord, because some codes do it differently.
The output of `get_bonds` is bidirectional. It uses doubly as much memory because it does not use the symmetry for storage compression, but makes it easier to write such expressions.
Namely you have:
```python3
j in bonds_atoms[i] == i in bonds_atoms[j]
```

src/quemb/kbe/fragment.py
mcocdawc: Yes, makes absolutely sense

src/quemb/molbe/chemfrag.py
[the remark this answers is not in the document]
mscho527: If I simplify the loop, for graphene BE2, I have
```
Atom indices of motifs (1-indexed)
------------------------------------------------
 Fragment |    Origin |    Centers |      Edges
------------------------------------------------
        1 |         1 |          1 |       2, 4
        2 |         2 |          2 |       1, 5
        3 |         3 |          3 |       2, 4
        4 |         4 |          4 |    1, 3, 7
        5 |         5 |          5 |    2, 6, 8
        6 |         6 |          6 |       1, 5
        7 |         7 |          7 |    4, 6, 8
        8 |         8 |          8 |    3, 5, 7
------------------------------------------------
```
whereas without the simplification, I get
```
------------------------------------------------
 Fragment |    Origin |    Centers |      Edges
------------------------------------------------
        1 |         1 |          1 |    2, 4, 6
        2 |         2 |          2 |    1, 3, 5
        3 |         3 |          3 |    2, 4, 8
        4 |         4 |          4 |    1, 3, 7
        5 |         5 |          5 |    2, 6, 8
        6 |         6 |          6 |    1, 5, 7
        7 |         7 |          7 |    4, 6, 8
        8 |         8 |          8 |    3, 5, 7
------------------------------------------------
```

This is periodic graphene, so each BE2 fragment should have four atoms (one origin and three edges)

src/quemb/molbe/chemfrag.py
```diff
@@ -666,7 +782,11 @@ def is_ordered(self) -> bool:
 
     def write_geom(self, prefix: str = "f", dir: Path = Path(".")) -> None:
         """Write the structures of the fragments to files."""
-        mol = Cartesian.from_pyscf(self.mol)
+        mol = (
+            Cartesian.from_pyscf(self.mol)
+            if isinstance(self.mol, Mole)
+            else Cartesian(self.mol.to_mol())
```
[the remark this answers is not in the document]
mcocdawc: It might be worth a thought to write the fragments including the supercell. But this can be postponed to another PR

src/quemb/molbe/chemfrag.py
```diff
@@ -951,7 +1071,7 @@ def _ensure_no_shared_centers(
     @classmethod
     def from_frag_structure(
```
mcocdawc, line 1072: I think the iao_valence_mol also has to be redeclared.

mcocdawc (approved):
Thank you very much! Looks great.


You also get a  happy duck now. :-)

```text
⠀⠀⠀⠀⠀⠀⠀⢀⣠⠶⠛⠛⠛⠛⠓⠶⣄⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⣴⠶⣤⣀⣀⣀⣤⡟⠁⣴⣖⢶⡀⠀⠀⠀⠈⠻⣆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⢻⡄⠀⠉⢹⡇⡾⠀⠘⢿⡿⠟⠀⠀⠀⠀⠀⠀⠘⣆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⢹⣷⣤⣀⣀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠙⠦⣤⣬⣩⡿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⡏⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠈⢧⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⠟⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⢸⠇⠀⠀⠀⠀⠀⠀⢀⣴⣞⠁⠀⠀⠀⠀⠀⠀⢠⡞⠛⢦⡀⠀
⠀⠀⠀⠀⢀⡴⠋⠀⠀⠀⠀⠀⠀⠀⠈⠉⠉⠛⠳⢦⣄⣀⢀⣠⠟⠀⠀⠀⢻⡄
⠀⠀⢀⡴⠋⠀⠀⠀⣀⣤⠶⠶⠶⢤⣀⠀⠀⠀⠀⠀⠀⠉⠉⠁⠀⠀⠀⠀⠀⣷
⠀⢀⡾⠁⠀⠀⢀⡾⠋⠀⠀⠀⠀⠀⠙⢧⣄⡀⠀⠀⣀⣠⣤⡆⠀⠀⠀⠀⠀⣿
⠀⣼⠁⠀⠀⠀⢸⡅⠀⠀⠀⠀⠀⠀⠀⠀⠉⠉⠛⠉⠉⠁⢸⡇⠀⠀⠀⠀⠀⣿
⠀⣿⠀⠀⠀⠀⢸⡆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⡟⠀⠀⠀⠀⠀⢠⡏
⠀⢹⡄⠀⠀⠀⠀⠻⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⡴⠋⠀⠀⠀⠀⠀⢠⡟⠀
⠀⠈⢷⡄⠀⠀⠀⠀⠈⠛⠶⢦⣤⣤⣤⣤⣤⠴⠞⠉⠀⠀⠀⠀⠀⠀⣰⠟⠀⠀
⠀⠀⠀⠙⢦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⣴⠞⠁⠀⠀⠀
⠀⠀⠀⠀⠀⠉⠛⠶⢤⣄⣀⣀⣀⣀⣀⣀⣀⣀⣀⣤⡤⠶⠛⠉⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠉⠉⠉⠉⠉⠉⠉⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```

PR #145, opened 2025-04-11
Update SCI solver

lweisburn:
Add functionality for the SCI fragment solver to the code, modifying and adding necessary keywords


lweisburn:
Here, I updated the SCI section of the code and added some examples and tests (which work only when the relevant SCI codes are installed and compiled, and as such are not run automatically with Github actions). I only worked on the SCI solver, which uses the `cornell-shci` off `shci` codebases, which I found the simplest SCI code to work with. The others are left as "TODO", as well as their relevant keywords, in case we want to implement this in the future.

src/quemb/molbe/solver.py
```diff
@@ -188,10 +189,11 @@ def from_user_input(cls, user_args: DMRG_ArgsUser, mf: RHF):
 
 @define(frozen=True)
 class SHCI_ArgsUser(UserSolverArgs):
-    hci_pt: Final[bool] = False
     hci_cutoff: Final[float] = 0.001
-    ci_coeff_cutoff: Final[float | None] = None
-    select_cutoff: Final[float | None] = None
+    hci_pt: Final[bool] = False
+    ci_coeff_cutoff: Final[float | None] = None  # TODO SOLVER
```
mcocdawc, line 195: As talked in person.

Personally, I would remove unused arguments. Unless you plan to implement them in the near future.

mcocdawc (approved):
cool cool

I would remove some of the unused components, but if you would like to keep them, feel free to go ahead

PR #147, opened 2025-04-15
relAO_per_origin + added swallow replace option

mcocdawc:
**Summary**
Renamed `centerf_idx` to `relAO_per_origin` to more accurately reflect its current role.

**Context**
Conceptually, matching conditions should be enforced for all centers, i.e., using the `relAO_per_center` component. However, at present, neither the fragmentation logic nor the matching conditions support this. Instead the edges are **only matched against origins** which becomes clear from the renaming to `relAO_per_origin`. While adding support for it in chemgen and graphgen would be straightforward, exposing `relAO_per_center` without updating the matching logic would lead to mismatched array shapes.

This inconsistency is likely the root cause of the convergence issues that @lweisburn observed when working with BE3 and matching conditions.

**Notes**
This bug only manifests in systems that include centers which are not origins. As such, it does not appear in systems without boundaries (e.g., graphene, fullerenes), which is why it wasn't consistently observed.

**Temporary Fix**
As a stopgap and to confirm the bug I introduced a `swallow_replace` option. This ensures that any swallowed/contained fragment is replaced by the largest containing fragment. While this incurs a performance penalty, it guarantees that all centers are origins and will be correctly matched.

Upon `BE(>=3) and matching conditions and (there are centers that are not origins)` the code errors out and recommends to use the `swallow_replace`  option.

**Follow-Up**
I'll open an issue to track updating the matching condition logic in tandem with fragmentation to fully support matching conditions in an efficient manner. (#150)


mscho527:
Just so I understand correctly, if `swallow_replace=True`, it won't swallow, right? And we would have additional matching constraints (edge to center that isn't origin)?

mcocdawc:
> Just so I understand correctly, if `swallow_replace=True`, it won't swallow, right? And we would have additional matching constraints (edge to center that isn't origin)?

It won't swallow **and** it will replace the smaller fragment with the largest containing fragment.
This is not the same as purely not swallowing.
Indeed, this will give you the additional matching conditions of edge to center that isn't origin.

I put this PR on hold for the moment, because after the discussion yesterday with Troy I will raise an exception 
if `BE(>=3) and matching conditions and (there are centers that are not origins)`.

This is much easier done after merging
https://github.com/troyvvgroup/quemb/pull/136
https://github.com/troyvvgroup/quemb/pull/137

mscho527 (approved):
looks good to me!

PR #151, opened 2025-04-22
Better init of pfrag

mcocdawc:
C_lo_eo -> TA_lo_eo

- [ ] do the same for kbe


mcocdawc:
This PR breaks UBE unfortunately. It is non-trivial to pull the `.sd` call into the constructor of pfrags. 
A more conservative approach is in #153

PR #153, opened 2025-04-23
Improved init of pfrag

mcocdawc:
- Consistent output of `relAO_per_edge_per_frag` even if there are no edges. It will now return a list of empty lists and not an empty list. This means one can just loop over the empty list of edges and not special case for the case of no edges being present.

- Enables a nicer initialisation of the pfrag objects.

- [x] Does it break UBE? I cannot test that myself @lweisburn . :-(


src/quemb/kbe/pfrag.py
```diff
@@ -174,8 +186,9 @@ def sd(
         else:
             raise ValueError(f"Imaginary density in Full SD {max_val}")
 
-        Sites = [i + (nlo * 0) for i in self.AO_per_frag]
-        if not frag_type == "autogen":
+        Sites = [i + (nlo * 0) for i in self.AO_in_frag]
+        # TODO: Discuss with Minsik
```
mcocdawc, line 190: the following sort might be relevant for #144 @mscho527

src/quemb/molbe/pfrag.py
```diff
@@ -141,28 +161,16 @@ def sd(self, lao, lmo, nocc, thr_bath, norb=None, return_orb_count=False):
...
-                lmo,
-                nocc,
-                self.AO_per_frag,
-                thr_bath=thr_bath,
-                norb=norb,
-                return_orb_count=return_orb_count,
-            )
-        else:
-            TA = schmidt_decomposition(lmo, nocc, self.AO_per_frag, thr_bath=thr_bath)
-        self.C_lo_eo = TA
```
mcocdawc, line 160: I assume this is a typo and they meant `TA_lo_eo` instead of `C_lo_eo`

src/quemb/kbe/pfrag.py
mcocdawc: It was already fixed 
https://github.com/troyvvgroup/quemb/pull/144/files#diff-059f900a6633877c395553bbd7e267247aef938a4d8f0b50f4494690a85178aeL172
:-D

src/quemb/molbe/autofrag.py
```diff
@@ -290,13 +306,14 @@ def remove_nonnunique_frags(self, natm: int) -> None:
 
     def to_FragPart(self, mol: Mole, n_BE: int, frozen_core: bool) -> FragPart:
         MISSING = []  # type: ignore[var-annotated]
+        MISSING_PER_FRAG = [[] for _ in range(len(self.AO_per_frag))]  # type: ignore[var-annotated]
```
mcocdawc, line 309: @ShaunWeatherly 
Previously, when there were no edges, `relAO_per_edge` was just a flat empty list rather than a list of empty lists per fragment. This led to a lot of special casing in downstream code.

I changed the output from an empty list to an empty list of empty lists also for `graphgen`.

lweisburn:
The UBE code checks reveal that this PR should be okay, but also that the newer versions of PySCF/numpy have issues when running the unrestricted calculations. These will have to be addressed in a later PR (raising issue #157 )

lweisburn:
>  Does it break UBE? I cannot test that myself @lweisburn . :-( 

The UBE tests show that this PR should be okay, but also that the newer versions of PySCF/numpy have issues when running the unrestricted calculations. These will have to be addressed in a later PR (raising issue #157 )

src/quemb/kbe/pfrag.py
```diff
@@ -34,19 +46,19 @@ class Frags:
...
-        relAO_in_ref_per_edge=None,
-        centerweight_and_relAO_per_center=None,
-        eri_file="eri_file.h5",
-        unitcell_nkpt=1,
-        ewald_ek=None,
-        relAO_per_origin=None,
-        unitcell=1,
-    ):
+        *,
+        AO_in_frag: Sequence[GlobalAOIdx],
```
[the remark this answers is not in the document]
mcocdawc: There were even a couple more discrepancies. Thank you for spotting and fixed.

src/quemb/kbe/fragment.py
```diff
@@ -118,6 +120,21 @@ def all_centers_are_origins(self) -> bool:
             )
         )
 
+    def to_Frags(self, I: int, eri_file: PathLike, unitcell_nkpt: int) -> Frags:
+        return Frags(
+            AO_in_frag=self.AO_per_frag[I],
```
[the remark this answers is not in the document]
mcocdawc: Already clarified in person.

> AO_in_frag are the AO indices in a particular fragment. (pfrags.Frags) It is of type Sequence[GlobalAOIdx]

> AO_per_frag are the AO indices per fragments, basically a list of AO_in_frag. (FragPart) It is of type Sequence[Sequence[GlobalAOIdx]]

PR #154, opened 2025-04-23
Making the relaxed and unrelaxed CCSD logic consistent

lweisburn:
In the current version of the code, the CCSD function sometimes will redundantly calculate the RDM2 (and needlessly calculate the wrong version). In this version, the RDM2 is calculated only when requested or when the energy is requested. 

This also disconnects the relaxed density and RDM return. These got linked somewhere along the line, but they shouldn't be. Now, the RDM returns are all connected to the `solve_ccsd` function, and the correct version is returned when choosing relaxed or unrelaxed densities. 

I also (unintentionally) am clarifying the RDM return for the other solvers to be more consistent. CCSD is still treated differently, with the option to return t-amplitudes only for memory's sake, but this can be changed in future PR's if desired.


mscho527 (approved):
looks good to me!

PR #156, opened 2025-04-23
Actually do Schmidt MO transformation

mcocdawc:
Implement Schmidt MO transformation for semi-sparse density fitting

This commit introduces the actual Schmidt molecular orbital (MO) transformation within the semi-sparse density fitting framework.

The natural benchmark for comparison is the integral-direct density fitting (DF) code. While the new implementation shows slightly slower timings for small molecules, it has significantly improved scaling behavior for larger systems.

For example, when increasing the size of a linear hydrocarbon chain from 48 to 80 carbon atoms, the timing comparison is as follows:
```
Integral-direct DF:
  0:59 → 6:55

Semi-sparse DF with Schmidt MO transformation:
  1:13 → 2:09
```

For the 80 carbon chain the energy difference to the integral direct DF results are 10^{-5} Hartree.


src/quemb/molbe/eri_sparse_DF.py
```diff
@@ -981,3 +1129,203 @@ def find_screening_radius(
...
+    write_eris(Fobjs, eris, file_eri_handler)
+
+
+def _slow_transform_sparse_DF_integral(
+    mf: scf.hf.SCF,
+    Fobjs: Sequence[Frags],
+    auxbasis: str | None = None,
+    screen_radius: Mapping[str, float] | None = None,
+) -> list[Matrix[np.float64]]:
+    """Only exist for reference. Can be deleted soon."""
```
mscho527, line 1211: I think it is useful to keep this :)

mscho527 (approved):
coool!

src/quemb/molbe/eri_sparse_DF.py
mcocdawc: in this case I perhaps will even expose it to the outside as `int_transfrom="slow-debug-sparse-DF"` and include it in one of the tests.

PR #158, opened 2025-04-25
Final fragpart rename

mcocdawc:
Final renaming of fragpart to keep consistent names everywhere.


src/quemb/kbe/autofrag.py
```diff
@@ -2349,20 +2349,26 @@ def autogen(
...
 
-        relAO_in_ref_per_edge.append(idx)
+        relAO_in_ref_per_edge_per_frag.append(idx)
 
-    if not AO_per_edge:
-        AO_per_edge = [[] for _ in range(n_frag)]
-        ref_frag_idx_per_edge = [[] for _ in range(n_frag)]
-        relAO_per_edge = [[] for _ in range(n_frag)]
-        relAO_in_ref_per_edge = [[] for _ in range(n_frag)]
+    if not AO_per_edge_per_frag:
```
lweisburn, line 2354: why is this section here twice?

src/quemb/kbe/pfrag.py
```diff
@@ -63,37 +63,34 @@ def __init__(
...
+        relAO_per_origin :
+            indices of the origin in the fragment, by default None
+        weight_and_relAO_per_center :
             weight used for energy contributions, by default None
-        eri_file : str, optional
+        eri_file :
             two-electron integrals stored as h5py file, by default 'eri_file.h5'
-        relAO_per_origin : list, optional
-            indices of the origin in the fragment, by default None
+        unitcell_nkpt:
```
lweisburn, line 91: do we have descriptions of these two keywords?

src/quemb/molbe/chemfrag.py
```diff
@@ -1267,13 +1267,11 @@ def _get_FragPart_no_iao(self) -> FragPart:
         # (which there is usually only one per fragment).
         relAO_per_origin = [
             union_of_seqs(*idx_per_origin)
-            for idx_per_origin in _extract_values(self.relAO_per_origin)
+            for idx_per_origin in _extract_values(self.relAO_per_origin_per_frag)
         ]
-        # A similar issue occurs for ``centerweight_and_relAO_per_center``
-        # where we have to form the union.
-        centerweight_and_relAO_per_center = [
+        weight_and_relAO_per_center_per_frag = [
```
lweisburn, line 1272: I see the warning was removed, is there any reason that this is no longer helpful?

src/quemb/molbe/helper.py
```diff
@@ -221,7 +221,7 @@ def get_frag_energy(
     mo_coeffs,
     nsocc,
     n_frag,
-    centerweight_and_relAO_per_center,
+    weight_and_relAO_per_center,
```
lweisburn, line 224: I like this shorter name!

lweisburn (approved):
looks good to me!

src/quemb/kbe/autofrag.py
mcocdawc: because I am stupid. thanks for catching!

src/quemb/molbe/chemfrag.py
mcocdawc: which warning do you mean?

mscho527: afaik, it's about the comment (" a similar issue ~") that was removed

mcocdawc: I just had the feeling it adds more clutter than value to the reader.

src/quemb/kbe/pfrag.py
mcocdawc: Postponed to #159; I did not add these arguments to the function, but just put it to the docstring.
Hence, I would proceed with this PR.

PR #162, opened 2025-05-01
Fix sphinx-multiversion whitelist

mscho527:
because i didn't expect `-alpha` back in the day 😅 

This allows `v0.1.0-alpha` to be recognized, not just `v0.1.0`

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -91,6 +91,6 @@ html_sidebars = {
 
 # -- Sphinx Multiversion --------------------------------------------------
 # https://holzhaus.github.io/sphinx-multiversion/master/configuration.html#
-smv_tag_whitelist = r"^v\d+\.\d+\.\d+.*$"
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -91,6 +91,6 @@
 
 # -- Sphinx Multiversion --------------------------------------------------
 # https://holzhaus.github.io/sphinx-multiversion/master/configuration.html#
-smv_tag_whitelist = r"^v\d+\.\d+\.\d+$"
+smv_tag_whitelist = r"^v\d+\.\d+\.\d+.*$"
```
mcocdawc, line 94: perhaps change the last .* to be a bit more specific.

```python
r"^v\d+\.\d+\.\d+(-[a-zA-Z0-9\.]+)?$"
```

mcocdawc (approved):

PR #166, opened 2025-05-14
UBE update

lweisburn:
Updating UBE to work with current versions of PySCF and numpy


src/quemb/scripts/__init__.py
lnorthcote, line 1: Not sure whether the HFCC script should be in this /scripts/ folder I made or somewhere else. Let me know if I should move it somewhere else.

src/quemb/.gitignore
lnorthcote, line 1: Also not sure about this, I just wanted to get rid of the diff. I made an example call file but that's not needed for everyone else so I can also move it to one of my own folders.

src/quemb/molbe/be_parallel.py
```diff
@@ -407,7 +407,7 @@ def run_solver_u(
         gcores=full_uhf.full_gcore,
         frozen=frozen,
     )
-    return e_f
+    return e_f, rdm1_tmp, fobj_a._mf.mo_coeff.copy(), fobj_b._mf.mo_coeff.copy()
```
mscho527, line 410: Note: Plan to bring consistent function return signature with #278 @cariluo

src/quemb/molbe/mbe.py
mscho527, line 1: Creating #283  to properly fix the mypy error in a future PR

src/quemb/molbe/ube.py
```diff
@@ -403,6 +418,54 @@ def oneshot(self, solver="UCCSD", nproc=1, ompnum=4):
...
+            Wb = self.W[1] if self.frozen_core else self.W
+            Proj_a = self.S @ Wa[:, cind] @ Wa[:, cind].T @ self.S
+            Proj_b = self.S @ Wb[:, cind] @ Wb[:, cind].T @ self.S
+
+            mca = get_mo(fobj_a)
+            mcb = get_mo(fobj_b)
+
+            # Transform fragment RDM to full AO space first
+            rdm1a_full = fobj_a.TA @ mca @ fobj_a.rdm1__ @ mca.T @ fobj_a.TA.T
+            rdm1b_full = fobj_b.TA @ mcb @ fobj_b.rdm1__ @ mcb.T @ fobj_b.TA.T
```
mscho527: I think we must use `fobj_a.mo_coeffs` (rather than selecting between candidates in `get_mo`) here, as we want the transformation matrix between the embedding orbitals and the fragment Hartree-Fock **(with BE potentials)** MOs here. I think `_mf` misses the potentials.

src/quemb/molbe/ube.py
```diff
@@ -45,6 +45,8 @@ def __init__(
         pop_method: str | None = None,
         compute_hf: bool = True,
         thr_bath: float = 1.0e-10,
+        equal_bath: bool = True,
+        use_df: bool = False,
```
mscho527: I think it is advantageous to match how integral transformation is specified in restricted BE. i.e., the density fitting option would be `int_transform: IntTransforms = "out-core-DF"` and `use_DF = False` would map to `int_transform: IntTransforms = "in-core"`, as we would probably want other int_transform options in restricted BE to be available with unrestricted BE in the future.

src/quemb/molbe/ube.py
```diff
@@ -200,52 +211,59 @@ def initialize(self, eri_, compute_hf):
...
+                        norb=fobj_b.n_b,
+                    )
+
+            assert fobj_a.TA is not None and fobj_b.TA is not None
+            if self.use_df:
+                eri_a = self.mf.with_df.ao2mo(fobj_a.TA, compact=True)
+                eri_b = self.mf.with_df.ao2mo(fobj_b.TA, compact=True)
+                eri_ab = self.mf.with_df.ao2mo(
+                    (fobj_a.TA, fobj_a.TA, fobj_b.TA, fobj_b.TA), compact=True
+                )
```
mscho527: If I follow correctly, this assumes `equal_bath == True`, right? I think there should be an assert to prevent late runtime errors with opposite spin ERI transform.

src/quemb/molbe/ube.py
```diff
@@ -200,52 +211,59 @@ def initialize(self, eri_, compute_hf):
...
             else:
-                eri_a = ao2mo.incore.full(
-                    eri_, fobj_a.TA, compact=True
-                )  # otherwise, do an incore ao2mo
+                assert eri_ is not None, "eri_ is None: set incore_anyway for UHF"
```
mscho527, line 283: This assert should probably way at the top to fail earlier than later

src/quemb/scripts/__init__.py
mscho527: I think this is fine; `example` probably also works, but I don't have a strong preference either way

tests/ube-oneshot_test.py
```diff
@@ -20,51 +26,55 @@ class TestOneShot_Unrestricted(unittest.TestCase):
         "This test is known to fail.",
     )
```
mscho527, line 240: Activate these tests?

tests/ube-oneshot_test.py
```diff
@@ -20,51 +26,55 @@ class TestOneShot_Unrestricted(unittest.TestCase):
...
         mol.charge = -1
         mol.spin = 1
         mol.build()
         self.molecular_unrestricted_oneshot_test(
-            mol, 1, "Hexene Anion Frz (BE1)", True, -0.35753374
+            mol, 1, "Hexene Anion Frz (BE1)", True, -0.35753375
         )
         self.molecular_unrestricted_oneshot_test(
-            mol, 2, "Hexene Anion Frz (BE2)", True, -0.34725961
+            mol, 2, "Hexene Anion Frz (BE2)", True, -0.34617685, delta=1e-4
```
mscho527, line 325: Did the value change due to equal_bath? If so, I think we should:

- Keep the old values in "Test A", use equal_bath = False, and keep KNOWN_TO_FAIL_TESTS pytest directive
- and Add a new test "Test B", where equal_bath = True, new values are used to compare, and BE3 (expensive) tests come with EXPENSIVE_TESTS pytest directive

mscho527 (changes requested):
Thanks for the fixes, Lea & Leah! I have some requested changes & questions, but great work!

src/quemb/molbe/ube.py
lnorthcote: Thanks! I think I introduced a `mo_coeff_uccsd` because I thought it was nice to keep track of methods implemented in UBE. But that doesn't match BE. So it was not using _mf as it always hit `mo_coeff_uccsd` first which was potential corrected. But I think you're right that the codes should match up nicely and the fallback thing is not needed. So I switched to `mo_coeffs` for UBE to mirror `be_func/be_func_parallel` and have tested that it uses the right matrix and simplified `get_mo`.

src/quemb/molbe/ube.py
lnorthcote: Yes, it does. I added an assert just before it now.

src/quemb/molbe/ube.py
lnorthcote: Yes, you’re right. I added an assert based on `fobj_a.TA.shape[1] != fobj_b.TA.shape[1]` for your other comment and added the ``eri_` assert just below that.

src/quemb/molbe/ube.py
lnorthcote: Yes, that makes sense. Now it's importing `IntTransforms` and using it the same way BE does. Currently only `"in-core"` and `"out-core-DF"` are supported in UBE, anything else raises `NotImplementedError` for now.

tests/ube-oneshot_test.py
lnorthcote: Yes, it did. I added two test sets now and let Test A keep the KNOWN_TO_FAIL_TESTS, thanks!

tests/ube-oneshot_test.py
lnorthcote: Activated Test B and kept Test A deactivated.

PR #170, opened 2025-06-26
Finalize Integral Transformation to Schmidt Orbitals

mcocdawc:
- Implements the semi-sparse DF in C++ + CUDA (8 times speedup on a 160 n_C over conventional density fitting)
- Bundle Eigen with our repo and switch to `pyproject.toml` for the build process

- keep a global TA matrix for fragment orbitals and have per-fragment offsets to index into this global TA matrix. This would allow to reuse info about shared fragment MOs (not used in the end in this PR, but might be relevant for other codes)

- changed many njit declarations into `nogil=True` so that we can have proper thread-based parallelism on top.

(followup in #171 )

- [x] fix test suite
- [x] pass on debug flag and do conditional print
- [x] go one last time over docstrings

I will delete the numba code in a follow up PR, but I want to have it once in the master branch and tag it there, so that we can go back and re-obtain it.

### DIFF
--- a/src/quemb/molbe/mbe.py
+++ b/src/quemb/molbe/mbe.py
@@ -1,7 +1,7 @@
 # Author(s): Oinam Romesh Meitei
 
 import pickle
-from typing import Literal, TypeAlias
+from typing import Final, Literal, TypeAlias
 from warnings import warn
 
 import h5py
@@ -102,7 +102,6 @@ class BE(MixinLocalize):
         auxbasis: str | None = None,
         MO_coeff_epsilon: float = 1e-4,
         AO_coeff_epsilon: float = 1e-10,
-        int_n_threads: int = 1,
     ) -> None:
         r"""
         Constructor for BE object.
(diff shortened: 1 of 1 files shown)


src/quemb/molbe/_cpp/eri_sparse_DF.cpp
```diff
@@ -0,0 +1,618 @@
...
+        : _n_rows(L_host.rows()), _n_cols(L_host.cols()), _size( _n_rows * _n_cols)
+    {
+        const size_t bytes = _size * sizeof(double);
+        CUDA_CHECK_THROW(cudaMalloc(reinterpret_cast<void **>(&d_L), bytes));
+        CUDA_CHECK_THROW(cudaMemcpy(d_L, L_host.data(), bytes, cudaMemcpyHostToDevice));
+    }
+
+    ~GPU_MatrixHandle()
+    {
+        if (d_L)
```
mcocdawc, line 53: why the hell do you use this class????!@!!!

src/quemb/molbe/_cpp/eri_sparse_DF.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+class SemiSparseSym3DTensor
+{
+  public:
+    explicit SemiSparseSym3DTensor(
+        Matrix unique_dense_data, std::tuple<int, int, int> shape, std::vector<std::vector<OrbitalIdx>> exch_reachable,
+        std::vector<std::vector<OrbitalIdx>> exch_reachable_unique,
+        std::vector<std::vector<std::pair<std::size_t, OrbitalIdx>>> exch_reachable_with_offsets,
+        std::vector<std::vector<std::pair<std::size_t, OrbitalIdx>>> exch_reachable_unique_with_offsets,
+        std::unordered_map<std::size_t, std::size_t> offsets)
```
mcocdawc, line 94: `std::map` is preferred over `std::unordered_map`

mcocdawc (commented):
looks overall good. but wtf

.vscode/c_cpp_properties.json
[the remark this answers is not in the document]
mcocdawc: good point. removed it

mscho527 (approved):
Great! I have one small request, but that's it. This PR will make the code much more useful. Thanks!

PR #176, opened 2025-07-10
Allow periodic chemgen to account for periodicity with bonds_atoms input

mscho527:
Allow periodic chemgen to account for periodicity for the `bonds_atoms` optional argument
Print out some additional message when `treat_H_different` runs into exception
Expose `modify_atom_data` in chemcoord to allow granular control over vdW radius for each atom


mscho527:
@mcocdawc @nwhel 
ready for review!

nwhel (approved):
These look good, but should the docstring for modify_atom_data be added to the rest of the methods (e.g. from_cartesian)? Otherwise it looks like all the new functionality is good.

mscho527:
> These look good, but should the docstring for modify_atom_data be added to the rest of the methods (e.g. from_cartesian)? Otherwise it looks like all the new functionality is good.

ah good point -- i'll add them
@mcocdawc now ready for your review!

src/quemb/molbe/chemfrag.py
```diff
@@ -196,6 +197,9 @@ def from_cartesian(
               to change the radius of individual elements, e.g. :python:`{"C": 1.5}`.
 
             The keyword is mutually exclusive with :python:`bonds_atoms`.
+        modify_atom_data :
+            To change the van der Waals radius of one or more specific atoms, pass a
+            dictionary that looks like {index1: 1.5}.
```
mcocdawc, line 202: I would use
```
:python:`{index1: 1.5}`
```
here and at other occurences, where it should be rendered as code

src/quemb/molbe/chemfrag.py
```diff
@@ -250,13 +255,28 @@ def motifs_share_H() -> bool:
...
+
         def all_H_belong_to_motif() -> bool:
             return H_atoms.issubset(union_of_seqs(*(H_per_motif.values())))
 
         if treat_H_different and not (all_H_belong_to_motif() and not motifs_share_H()):
             raise ValueError(
                 "Cannot treat hydrogens differently if not all hydrogens belong "
-                "to exactly one motif."
+                "to exactly one motif.\nH not contained in any motif: "
+                f"{list(H_atoms.difference(union_of_seqs(*(H_per_motif.values()))))}"
```
mcocdawc, line 275: I don't think it's necessary to cast to a list.

mcocdawc (approved):
The code looks great but a test in `test_chemfrag` would be nice.

Issue #172, opened 2025-07-01
Modernize to >C++20


mcocdawc:
Currently we cannot use C++20, because telemachus only has gcc 8.3

Eventually we want to change to concepts and auto in template functions.

e.g.
```C++
// Sum of integers from 1 to n
template <typename T> constexpr int_t gauss_sum(const T n) noexcept {
    const int_t N = to_int_t(n);
    return (N * (N + 1)) / 2;
}
```
becomes
```C++
// Sum of integers from 1 to n
constexpr int_t gauss_sum(std::integral auto n) noexcept {
    const int_t N = to_int_t(n);
    return (N * (N + 1)) / 2;
}
```

mcocdawc:
Currently, we could not use structured bindings in combination with OpenMP and clang.
If clang ever fixes https://github.com/llvm/llvm-project/issues/33025

then revert the change in https://github.com/troyvvgroup/quemb/pull/182/files/6546bad79245a6341a5cb5b0baa315ad507a24e0#diff-fdd7421e5d69c98a881e49fd7840658f4f3dd5a5ded26a964eede3978d8d69bbL402
to use
```C++
const auto [naux, nao, nmo] = int_mu_i_P.get_shape();
```
instead

PR #182, opened 2025-07-23
Support more compilers

mcocdawc:
- Support GCC, Clang, Intel, and Microsoft C++ compilers

- both GCC and clang are now tested in debug mode as part of the test suite

### DIFF
--- a/src/quemb/molbe/_cpp/eri_sparse_DF.cpp
+++ b/src/quemb/molbe/_cpp/eri_sparse_DF.cpp
@@ -16,17 +16,21 @@
 #include <cuda_runtime_api.h>
 #endif
 
+#ifdef GCC
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Warray-bounds"
 #pragma GCC diagnostic ignored "-Wsign-conversion"
 #pragma GCC diagnostic ignored "-Wstringop-overread"
 #pragma GCC diagnostic ignored "-Wnull-dereference"
+#endif
 #include <pybind11/eigen.h>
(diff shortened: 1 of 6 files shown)


src/quemb/molbe/_cpp/eri_sparse_DF.cpp
```diff
@@ -398,13 +402,22 @@ Tensor3D copy_from_numpy(py::array_t<double, py::array::f_style> arr)
 Matrix contract_with_TA_2nd_to_sym_dense(const SemiSparse3DTensor &int_mu_i_P, const Matrix &TA) noexcept
 {
     PROFILE_FUNCTION();
+#ifndef CLANG
     const auto [naux, nao, nmo] = int_mu_i_P.get_shape();
-
+#else
+    // Clang does not yet support capturing structured bindings in OpenMP.
```
lweisburn, line 408: very helpful change haha

mcocdawc: I set up a reminder for the future

https://github.com/troyvvgroup/quemb/issues/172#issuecomment-3114127058

mcocdawc:
> Helpful changes, thank you! This now works on my local computer and supports clang, not just gcc! I can add a little bit to the wiki for mac installation, because the default clang on macs does not work with open_mp. I'll find and link the help pages for it.
> 
> I do think though it would be helpful to update the `README`--the code does not install (at least for me) with the same `git clone <code>` command anymore, it requires the recursive download of submodules.

Good point.
I updated the README and `install.rst` to explicitly list `--recurse-submodules`.

PR #185, opened 2025-07-23
turn timing into a decorator

### LINKED ISSUE #183: Turn the timing of functions into a decorator
> Yes, the logging library seems like a good idea
> 
> Actually the first thing to do is far less technical, but just a decision on our conventions, similar to https://docs.python.org/3/library/logging.html#logging-levels
> 
> So far different people (including me) added if statements of the form
> ```python3
> if (PRINT_LEVEL >= n)
> ```
> where it was never clear, what the different `n`s would mean.
> Already just giving them a proper name as in the attached table would improve it a lot.
> 
> 
> ---------------------------------------
> 
> Some more thoughts. Basically we do logging for two large purposes.
> 
> 1. Report more detailed info, also on memory usage and timings. This is useful for both users and developers.
> 2. Log values deep in the call chain. This is mostly relevant when debugging and more for developers.
> 
> The lowest log level should only contain these timings and memory usage information to not bloat the output with debugging info, when it is not needed.
> 
> Because of this I would also be fine with just three logging levels. 
> 
> - None
> - Timings, Memory usage
> - Debugging logs
> 
> ---------------------------------------------------
> 
> Finally, it would be great to turn the timing of functions into a decorator.
> 
> 
> ```python3
> import time
> import functools
> from collections import defaultdict
> from attrs import define, field
> 
> @define
> class FunctionTimer:
>     stats: dict = field(factory=lambda: defaultdict(lambda: {'time': 0.0, 'calls': 0}))
> 
>     def timeit(self, func):
>         """Decorator to time a function and record stats."""
>         @functools.wraps(func)
>         def wrapper(*args, **kwargs):
>             start = time.perf_counter()
>             result = func(*args, **kwargs)
>             duration = time.perf_counter() - start
>             key = f"{func.__module__}.{func.__qualname__}"
>             self.stats[key]['time'] += duration
>             self.stats[key]['calls'] += 1
>             return result
>         return wrapper
> 
>     def print_top(self, n=10):
>         """Print the top-n functions by total accumulated time."""
>         sorted_stats = sorted(
>             self.stats.items(),
>             key=lambda item: item[1]['time'],
>             reverse=True
>         )
>         print(f"{'Function':60} {'Calls':>10} {'Total Time (s)':>15}")
>         print('-' * 90)
>         for name, data in sorted_stats[:n]:
>             print(f"{name:60} {data['calls']:10d} {data['time']:15.6f}")
> ```
> which you would use as:
> 
> 
> ```python3
> timer = FunctionTimer()
> 
> @timer.timeit
> def task_slow():
>     time.sleep(0.05)
> 
> @timer.timeit
> def task_fast():
>     time.sleep(0.01)
> 
> for _ in range(10):
>     task_slow()
> for _ in range(30):
>     task_fast()
> 
> timer.print_top(5)
> ```
> 
> which yields
> 
> ```
> Function                                                          Calls  Total Time (s)
> ------------------------------------------------------------------------------------------
> __main__.task_slow                                                   10        0.501382
> __main__.task_fast                                                   30        0.304721
> ```
> 
> Something similar for logging memory usage would be quite practical as well. 

 _Originally posted by @mcocdawc in [#25](https://github.com/troyvvgroup/quemb/issues/25#issuecomment-3039360427)_

cariluo:
closes #183

### DIFF
--- a/src/quemb/molbe/be_parallel.py
+++ b/src/quemb/molbe/be_parallel.py
@@ -121,10 +121,10 @@ def run_solver(
 
     # Select solver
     if solver == "MP2":
-        mc_ = solve_mp2(mf_, mo_energy=mf_.mo_energy)
-        rdm1_tmp = mc_.make_rdm1()
+        mc_mp2 = solve_mp2(mf_, mo_energy=mf_.mo_energy)
+        rdm1_tmp = mc_mp2.make_rdm1()
         if eeval:
-            rdm2s = mc_.make_rdm2()
+            rdm2s = mc_mp2.make_rdm2()
 
     elif solver == "CCSD":
         if eeval:
@@ -147,10 +147,10 @@ def run_solver(
             )
 
     elif solver == "FCI":
-        mc_ = fci.FCI(mf_, mf_.mo_coeff)
-        efci, civec = mc_.kernel()
+        mc_fci = fci.FCI(mf_, mf_.mo_coeff)
+        efci, civec = mc_fci.kernel()
         unused(efci)
-        rdm1_tmp = mc_.make_rdm1(civec, mc_.norb, mc_.nelec)
+        rdm1_tmp = mc_fci.make_rdm1(civec, mc_fci.norb, mc_fci.nelec)
 
     elif solver == "HCI":  # TODO
         # pylint: disable-next=E0611
@@ -265,7 +265,7 @@ def run_solver(
     if eeval:
         if solver == "FCI" or solver == "SCI":
             if solver == "FCI":
-                rdm2s = mc_.make_rdm2(civec, mc_.norb, mc_.nelec)
+                rdm2s = mc_fci.make_rdm2(civec, mc_fci.norb, mc_fci.nelec)
             if use_cumulant:
                 hf_dm = zeros_like(rdm1_tmp)
                 hf_dm[diag_indices(nocc)] += 2.0


cariluo:
Note this PR is for #185

cariluo:
Comment: I made a messy merge because I didn't pull changes from remote first. Lesson learned, will try to avoid this messiness in future.

cariluo:
There is an error with using `time.perf_counter` in one of the test files. I deliberately have left the import of time untouched.

src/quemb/shared/helper.py
```diff
@@ -119,18 +121,51 @@ def delete_multiple_files(*args: Iterable[Path]) -> None:
...
+timer = FunctionTimer()
+
+
 @define(frozen=True)
 class Timer:
     """Simple class to time code execution"""
 
     message: str = "Elapsed time"
-    start: float = field(init=False, factory=time)
+    start: float = field(init=False, factory=lambda: time.perf_counter())
```
mcocdawc, line 162:
```python3
lambda: time.perf_counter()
# is equivalent to
time.perfcounter
```

mcocdawc (commented):
Looks good, except a very tiny comment (see above).

One remaining question is how to do the printing in the end.
It would be great if you either
1. Write a docstring for the `FunctionTimer` class that emphasizes to the `user` that they have to call `timer.print_top()` at the end, if they want to see timings.

or

2. We actually register `timer.print_top` to be called at exit, via the [atexit module](https://docs.python.org/3/library/atexit.html).

I am more or less undecided between the two options. Perhaps you @cariluo or @mscho527 have a stronger opinion in any direction?

PS: I saw that you correctly replaced `from attr` with `from attrs`. Could you check if this was a problem somewhere else as well?

src/quemb/molbe/mbe.py
```diff
@@ -980,7 +978,6 @@ def initialize(
                 assert_never(int_transform)
         else:
             eri = None
-        logger.info(f"ERI transform time: {eritransform_timer.str_elapsed()}")
```
mscho527, line 983: To be clearer on my last comment, I would prefer it if things like this are still preserved. i.e., after ERI transformation, logger should print out how long it took. This is a metric we are often interested in.

mcocdawc: this could actually be added into the `timer.timeit` method. Then the developer does not have the deal with the clutter of logging statements.

src/quemb/shared/helper.py
cariluo: actually, because lambda is used, I actually need to do the `perf_counter()` version otherwise I get the error (see the failed check for the `fixed typo` commit)

mscho527 (approved):
Okay thanks for the changes! I approve this PR seeing that my concern was (partially) addressed. I think we still want to do some further work, to isolate each part of `initialize` for timing benchmark. i.e., we want timings for integral transformation (which we mostly care about), fock build, etc separately. This would likely require shelling out each routine to separate functions. Given how these are large changes, we could try to address them in a later PR.

PR #190, opened 2025-07-31
replace prints with warnings

### LINKED ISSUE #127: change some prints into warnings
See discussion in #126

cariluo:
closes #127

### DIFF
--- a/src/quemb/molbe/opt.py
+++ b/src/quemb/molbe/opt.py
@@ -207,8 +207,6 @@ class BEOPT:
                         )
                         break
                 if self.err >= self.conv_tol:
-                    warnings.warn(
-                        f"BE DID NOT CONVERGE IN {self.max_space} STEPS", stacklevel=2
-                    )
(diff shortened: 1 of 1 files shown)


cariluo:
Note: could also add categories, something like 

```
warnings.warn(
    f"BE DID NOT CONVERGE IN {self.max_space} STEPS",
    category=RuntimeWarning,
    stacklevel=2
)
```

Use stacklevel=2 (or higher) when you want the warning to point to user code rather than internal helper functions. 

The options for the categories are 
Screenshot 2025-07-31 at 9 30 42 AM

cariluo:
What other prints should I replace with warnings?

cariluo:
I think I can go ahead and push this one unless there's something wrong, and I am always happy to replace similar things with warning as well as people notice things.

mcocdawc (approved):
Nice small change!

Is there a specific reason, why you want the stacklevel=2 option here? IMHO the default, i.e. showing the trace to where the warning was raised, makes more sense (here).

PR #192, opened 2025-08-02
break apart initialize

### LINKED ISSUE #189: break apart initialize (for timing purposes)
We want to isolate each part of `initialize` for timing benchmark. i.e., we want timings for integral transformation (which we mostly care about), fock build, etc separately. This would likely require shelling out each routine to separate functions. 

_Originally posted by @mscho527 in https://github.com/troyvvgroup/quemb/pull/185#pullrequestreview-3072026580_

cariluo:
closes #189


cariluo:
@lweisburn or @mscho527 what is happening here? 
```
couti = 0
        for fobj in self.Fobjs:
            fobj.udim = couti
            couti = fobj.set_udim(couti)
```

cariluo:
What other changes would you guys like to see? I've pulled out the eri transform and the fragment processing so that they can be timed, but there are some bits and pieces at the beginning and end of initialize that I'm not sure how to incorporate.

src/quemb/molbe/mbe.py
```diff
@@ -851,6 +851,127 @@ def print_ini(self):
         print("-----------------------------------------------------------", flush=True)
         print(flush=True)
 
+    @timer.timeit
+    def eri_transform(self, int_transform, eri_, file_eri):
```
mcocdawc, line 855: If you are refactoring anyway, I think it would be great to add type hints.

src/quemb/molbe/mbe.py
```diff
@@ -851,6 +851,127 @@ def print_ini(self):
...
+                file_eri_handler=file_eri,
+                MO_coeff_epsilon=self.MO_coeff_epsilon,
+                AO_coeff_epsilon=self.AO_coeff_epsilon,
+                n_threads=self.n_threads_integral_transform,
+            )
+        else:
+            assert_never(int_transform)
+
+    @timer.timeit
+    def process_fragments(self, file_eri, restart, compute_hf):
```
mcocdawc, line 939: 1. Is it always called with `compute_hf=True`, if yes I would get rid of this boolean flag.

2. Do you have an idea for a more descriptive name than `process_fragments`? That's a name that basically applies to everything we do in BE. :-D 

3. It might make sense to prepend the new methods with `_`, i.e. declare them as private (by convention).

cariluo: I'm not sure what would be better than `process_fragments` because there's a lot that gets done. The docstring for your convenient reference is
```
Processes all molecular fragments by constructing their Fock matrices,
        performing SCF, and optionally computing fragment Hartree–Fock (HF) energies.

        This includes:

        - Loading and transforming ERIs (in-core or restored format).
        - Constructing 1-electron Hamiltonians via basis transformations.
        - Running fragment-level SCF calculations.
        - Building initial density matrices.
        - Computing and accumulating fragment HF energies if ``compute_hf`` is True.
        - Verifying HF-in-HF energy consistency.

        Parameters
        ----------
        file_eri : h5py.File
            HDF5 file containing fragment ERIs.
        restart : bool
            If True, skips ERI transformation and file closure.
        compute_hf : bool
            If True, computes fragment HF energies and compares total
            against the full system HF energy.
```

maybe `initialize_fragments`?

cariluo:
- [ ] Get rid of `compute_hf` check
- [ ] Do the same factorization that I've done for `mbe.py initialize` for `ube.py initialize`

src/quemb/molbe/mbe.py
```diff
@@ -68,6 +68,17 @@
...
+
+Options:
+- "in-core"
+- "out-core-DF"
+- "int-direct-DF"
+- "sparse-DF-cpp"
+- "sparse-DF-nb"
+- "sparse-DF-cpp-gpu"
+- "sparse-DF-nb-gpu"
+"""
```
mscho527, line 81: Nitpick: I don't think it is necessary to repeat the options here as comment

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
+        and stores them in a file.
+
+        Transformation strategy follows a decision tree:
+
+        1. If full (ij|kl) ERIs are available:
+           - Use in-core ``ao2mo`` transformation.
+        2. Else, if (ij|P) intermediates from density fitting are available:
+           - Use out-of-core ``ao2mo`` transformation with saved (ij|P).
+        3. Else, if ``integral_direct_DF`` is requested:
+           - Use on-the-fly density-fitting integral evaluation.
```
mscho527, line 912: Decision tree is old (does not contain the sparse options); we should update

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
 
         Parameters
         ----------
+        int_transform : :py:data:`~quemb.molbe.mbe.IntTransforms`
```
mscho527, line 916: @mcocdawc are type specifications in the docstrings redundant here?

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
 
     @timer.timeit
-    def initialize(
-        self,
-        eri_,
-        compute_hf: bool,
-        *,
-        restart: bool,
+    def _eri_transform(
+        self: "BE",
```
mscho527, line 896: not entirely sure what `"BE"` here is for?

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
-        self.all_fragment_MO_TA, frag_TA_index_per_frag = union_of_frag_MOs_and_index(
-            self.Fobjs, self.mf.mol.intor("int1e_ovlp"), epsilon=1e-10
-        )
-        for fobj, frag_TA_offset in zip(self.Fobjs, frag_TA_index_per_frag):
-            fobj.frag_TA_offset = frag_TA_offset
+    @timer.timeit
+    def _process_fragments(self: "BE", file_eri: h5py.File, restart: bool):
+        """
+        Processes all molecular fragments by constructing their Fock matrices,
+        performing SCF, and optionally computing fragment Hartree–Fock (HF) energies.
```
mscho527, line 998: no longer optional after we removed compute_hf, right?

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
+        else:
+            assert_never(int_transform)
 
-        self.all_fragment_MO_TA, frag_TA_index_per_frag = union_of_frag_MOs_and_index(
-            self.Fobjs, self.mf.mol.intor("int1e_ovlp"), epsilon=1e-10
-        )
-        for fobj, frag_TA_offset in zip(self.Fobjs, frag_TA_index_per_frag):
-            fobj.frag_TA_offset = frag_TA_offset
+    @timer.timeit
+    def _process_fragments(self: "BE", file_eri: h5py.File, restart: bool):
```
mscho527, line 995: same comment as above about `self` type annotation

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
-                    MO_coeff_epsilon=self.MO_coeff_epsilon,
-                    AO_coeff_epsilon=self.AO_coeff_epsilon,
-                    n_threads=self.n_threads_integral_transform,
-                )
-                eri = None
-            elif int_transform == "sparse-DF-nb-gpu":
-                from quemb.molbe.eri_sparse_DF import (  # noqa: PLC0415
-                    transform_sparse_DF_integral_nb_gpu,
-                )
+        - Loading and transforming ERIs (in-core or restored format).
```
mscho527, line 1002: This is misleading as we don't do basis transformation here. we simply change between different symmetry for the eri

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
+        - Constructing 1-electron Hamiltonians via basis transformations.
+        - Running fragment-level SCF calculations.
+        - Building initial density matrices.
```
mscho527, line 1005: dm0, while used as initial dm for subsequent scf (with BE potentials), are converged HF 1-RDM (without BE potentials) in MO basis

src/quemb/molbe/mbe.py
```diff
@@ -1039,19 +1036,55 @@ def initialize(
...
+        eri_ : numpy.ndarray
+            Electron repulsion integrals.
+        restart : bool, optional
+            Whether to restart from a previous calculation, by default False.
+        int_transfrom :
+            Which integral transformation to perform.
+        """
+        # Create a file to store ERIs
+        if not restart:
+            file_eri = h5py.File(self.eri_file, "w")
```
mscho527, line 1071: I don't know what the best practice would be for the `file_eri` bit here. @mcocdawc Should we both open and close `file_eri` in `initialize` and hand off `eri` to `_eri_transform` and `_process_fragments`? Or should we let both functions read and close `file_eri` within each function? (in which case eri will be read twice, but fragment ERIs are small)

To me, opening `file_eri` here and closing it in a different function feels a bit odd.

src/quemb/molbe/ube.py
```diff
@@ -171,6 +171,64 @@ def __init__(
 
         self.initialize(mf._eri, compute_hf)
 
+    @timer.timeit
```
mscho527, line 174: I don't think this needs to be timed

src/quemb/molbe/ube.py
```diff
@@ -171,6 +171,64 @@ def __init__(
...
+                    orb_count_b[I][0] + orb_count_b[I][1],
+                ),
+                flush=True,
+            )
+        print(
+            "____________________________________________________________________",
+            flush=True,
+        )
+
+    @timer.timeit
```
mscho527, line 209: Likewise, no need to time simple arithmetic

src/quemb/molbe/ube.py
```diff
@@ -171,6 +171,64 @@ def __init__(
...
+        hf_err = self.hf_etot - (E_hf + self.enuc + self.E_core)
+
+        self.ebe_hf = E_hf + self.enuc + self.E_core - self.ek
+        print(f"HF-in-HF error                 :  {hf_err:>.4e} Ha")
+        if abs(hf_err) > 1.0e-5:
+            warn("Large HF-in-HF energy error")
+            print("eh1 ", EH1)
+            print("ecoul ", ECOUL)
+
+    @timer.timeit
```
mscho527, line 220: Same as above; no need to time counting integers

src/quemb/molbe/ube.py
mscho527, line 1: Changes here might clash with #166. I would proceed with caution or delay refactoring to happen after it.

mscho527 (changes requested):
Thanks for the work! Some questions and requests...

src/quemb/molbe/mbe.py
mcocdawc: > in which case eri will be read twice, but fragment ERIs are small

That might seem like a nitpicky comment, but merely opening the file and creating the file handler does not read it.
Yes, there is some overhead to opening and closing, but opening and closing a few times is not too bad IMHO if it enables more encapsulation.
I would still aim to reduce the amounts of opening and closing, but that is a low-priority goal.

Where the question of opening and closing really makes a difference, is when it allows us to use a contextmanager, i.e.
```python3
with h5py.File(...) as file_eri_handler:
   .
```
I am very much for a contextmanager and if that requires multiple opens and closes, then so be it.

In that context, I actually liked the previous name `file_eri_handler` better. `self.eri_file` is a path while the file handler is, well, a file handler; the names make it explicit.

src/quemb/molbe/mbe.py
```diff
@@ -886,139 +892,130 @@ def print_ini(self):
...
-        self,
-        eri_,
-        compute_hf: bool,
-        *,
-        restart: bool,
+    def _eri_transform(
+        self: "BE",
         int_transform: IntTransforms,
-    ) -> None:
+        eri_: numpy.ndarray,
```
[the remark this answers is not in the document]
mcocdawc: and it can be declared using the `quemb.shared.typing` specifications as `Matrix[np.floating]`, so in total:
`eri_: Matrix[np.floating] | None`

src/quemb/molbe/mbe.py
mscho527: As is, it doesn't read twice. I am mostly uneasy about file handler being created in one function and closed in another. Maintenance-wise, I worry that this may lead us to forget to close the files while modifying the code in the future.

Context manager would be nice too. That would resolve the concern I have as well.

src/quemb/molbe/mbe.py
mcocdawc: > I can definitely declare the new methods as private. For my own knowledge, and so that I have the answer in writing, generally when would I want to declare a method as private?

It depends :-D.

If you know for sure that a method is meant to be used by users, then  you declare it to be public and (try to) guarantee to not change it in the future anymore and you try to make it pure and without implicit contracts.
For example, after a BE instance is initialized you can simply call `oneshot` and you don't have to worry about calling something else first.
You can also call it repeatedly, in that sense it is idempotent. 
The user can also decide to not call `oneshot`, the object will still be in a valid state.

On the other hand; basically all these initialization functions have an implicit order in which they have to be called. If some of them are not called then your object is not in a well-defined state. You can also not call them twice (at least I am not sure about it). You can also not decide to not call them.
So long story short. There is not much advantage for the user to manually call parts of the initialization routines, but they can easily shoot themselves in the foot.

There is no mechanism in python that prevents people from calling underscored (private) functions, but we just indicate. "If you do this, do this at your own risk and know what you do. We make no guarantees that we will change this API in the future and you may shoot yourself in the foot now".

src/quemb/molbe/mbe.py
mscho527: Note that prior to this PR, `initialize` both opened and closed `file_eri`.
With this PR, `initialize` opens `file_eri` and `process_fragments` closes it.

src/quemb/molbe/ube.py
mcocdawc: It is easier anway if they are separate. Hence, I would postpone `kbe` and `ube` refactoring to another PR.
(Unless they can not be separately changed.)

src/quemb/molbe/mbe.py
mcocdawc: > Note that prior to this PR, initialize both opened and closed file_eri.
With this PR, initialize opens file_eri and process_fragments closes it.

Yes, here I am fully seconding Minsik with the old C-rule: the scope that acquires a resource is the scope that closes it.

(Unless you create the file handler and explicitly return it from a function; in that case the calling scope would close it. But in a way the calling scope acquires the resource via the called function, so it's not a contradiction to the previous rule.)

src/quemb/molbe/mbe.py
cariluo: I am a bit confused on this point. I can remove `BE` but do I need to do `from __future__ import annotations`?

src/quemb/molbe/ube.py
cariluo: No problem, I can postpone `kbe` and `ube` refactoring to another PR.

cariluo:
It is one of life's great mysteries to me where `file_eri` is defined for the case where `restart=True`:
```
if not restart:
            file_eri = h5py.File(self.eri_file, "w")
            self._eri_transform(int_transform, eri_, file_eri)

        self._initialize_fragments(file_eri, restart)

        if not restart:
            file_eri.close()
```

cariluo:
@mcocdawc @mscho527 there are three unresolved things I'd like input on above, and then I can make (maybe final?) changes to this PR

src/quemb/molbe/mbe.py
mcocdawc: > I am a bit confused on this point. I can remove BE but do I need to do from __future__ import annotations?

Sorry for not making it clear.
In this particular situation you can just remove BE and you don't need `from __future__ import annotations`.

However, there are other situations, where you want do use a type in an interface such as `def f(x: T)` where `T` is not yet defined. Then you could do "T" or `from __future__ import annotations` and in this case the latter is preferred.

mcocdawc:
> It is one of life's great mysteries to me where `file_eri` is defined for the case where `restart=True`:
> 
> ```
>         if not restart:
>             file_eri = h5py.File(self.eri_file, "w")
>             self._eri_transform(int_transform, eri_, file_eri)
> 
>         self._initialize_fragments(file_eri, restart)
> 
>         if not restart:
>             file_eri.close()
> ```
> 
> ??? I need to understand this before I can use the context manager

Honestly, the `restart=True` code looks iffy to me.
Did anyone use it at all in recent time @mscho527 ?

mscho527:
> > It is one of life's great mysteries to me where `file_eri` is defined for the case where `restart=True`:
> > ```
> >         if not restart:
> >             file_eri = h5py.File(self.eri_file, "w")
> >             self._eri_transform(int_transform, eri_, file_eri)
> > 
> >         self._initialize_fragments(file_eri, restart)
> > 
> >         if not restart:
> >             file_eri.close()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ??? I need to understand this before I can use the context manager
> 
> Honestly, the `restart=True` code looks iffy to me. Did anyone use it at all in recent time @mscho527 ?

If it is a restart, it wouldn't need to do basis transformation and save the results in file_eri, I think?

PR #200, opened 2025-08-11
make clang work on Mac

cariluo:


cariluo:
So I don't know how to make it work automatically (because I don't understand the CMakeLists.txt file), but for now here is a pretty simple fix that works for me:
Screenshot 2025-08-11 at 12 55 15 PM

mscho527:
i don't own Mac nor do i understand how they do things so i may be wrong. does the "automatized" fix require installing stuff on the users' mac (which seems to be the case based on the wiki)? if so, i would just consider those as dependencies and ask the users to follow the steps, rather than installing them for the users.

README.md
```diff
@@ -26,8 +26,6 @@ processors.
 Comprehensive documentation for QuEmb is available online [here](https://vanvoorhisgroup.mit.edu/quemb/)
 and provides detailed infomation on installation, usage, API reference, and examples.
 
-[Download the PDF documentation](docs/source/_static/quemb.pdf)
-
```
cariluo, line 30: In a different PR this got moved to the index.rst file

lweisburn (approved):

PR #201, opened 2025-08-12
Fix: maintain consistency for HF reference (w/ or w/o BE potential)

mscho527:
As discussed in subgroup meetings, we want to maintain consistency in the energy expression such that we consistently use HF 1-RDM calculated with BE potential or w/o BE potential as appropriate.


mscho527:
As discovered by @lweisburn, this inconsistency leads to a higher discrepancy for CCSD (although FCI didn't). Now setting this as high priority

PR #209, opened 2025-08-19
small fix to the expensive unittest test result artifact file name

mscho527:
a tinyyy fix to make sure that the junit test results have different name for the expensive test

fixes the following error:
```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```


mscho527:
whop lol didn't notice that this was fixed in the last pr!

PR #210, opened 2025-08-19
Expose Calculator.input.memory in orca interface

mscho527:
This sets the memory (per core in MB) for orca job (i.e., it sets `%MaxCore` directive). This is necessary for large jobs where the default value is insufficient to hold intermediate data for Fock matrix diagonalization.


mcocdawc (approved):
Nice improvement. Thank you!

Perhaps, I would change the variable name to `memory_MB`; then it becomes more self-documenting.

PR #211, opened 2025-08-19
Fix workdir cleanup

mcocdawc:
- fixed the behaviour of `atexit.register` for our purposes. The function `register_clean_exit` executes a function at exit **only if no exception was raised**.

- Introduce a make_subdir method for the WorkDir where the subdirectory inherits the cleanup settings of the root dir.


mscho527 (approved):
looks good!

PR #215, opened 2025-08-21
update references

cariluo:


cariluo:
I will come back to this PR if it becomes higher priority. Otherwise I will close it for now and welcome anyone else to take over because I do not want to cause any unintended disruptions to higher priority changes.

PR #217, opened 2025-08-22
Adding CNOs to QuEmb

lweisburn:
This adds cluster natural orbitals to QuEmb. As outlined in an in-progress manuscript, CNOs are natural orbitals which can be generated and augmented to each fragment space. These are especially valuable in larger basis set calculations on limited computational resources. This PR includes different schemes to select CNOs.

This is in progress. This first implementation is naive and expensive (but functional for smaller systems!). In ongoing work, borrowing from DLPNO ideas, the integral transformations can be sped up. Additional CNO schemes may also be added.


lweisburn:
I merged Oskar's in-progress #222 into this to clean up the geometry data being fed into the CNO code: I think I messed up the git history on it a little bit (sorry Oskar), but it should hopefully be okay once that PR is merged first!

PR #220, opened 2025-09-08
added re_eval_HF

mcocdawc:
Added an option to re-evaluate the global Hartree Fock energy from MO coefficients to determine the HF-in-HF error.


mscho527 (approved):

cariluo (commented):
looks good, thanks for the in-person explanation

PR #221, opened 2025-09-10
print memory of (ij | P) already at info logging level

mcocdawc:
Crucial information was missing, because the logging print level was too low.


cariluo:
Only changes were to the standard output text and changing debug to info. Looks fine to me.

PR #222, opened 2025-09-16
expose chemgen info

mcocdawc:
Expose additional chemgen info for downstream code.

This allows to specialize via:
```python
if isinstance(x, ChemFragPart):
    x.fragmented....
```
and use the additional available info.


lweisburn:
With this change, I am able to use the structure information necessary for the CNO code from the updated `FragPart` object. Works for me!

PR #225, opened 2025-10-07
try pathos multiprocessing

mcocdawc:
Use pathos multiprocessing as drop-in for `multiprocessing.Pool`.

If this does not fix the pickling issues we shall try `joblib`.


mscho527 (approved):
I think this is good to go, after testing this on the cluster as a batch job!

PR #236, opened 2025-10-17
Remove numba sparse DF

mcocdawc:
Removed numba implementation of sparse DF.

This simplified the code as it removes a lot of unused functionality.

Furthermore we can use the C++ objects exposed to python now more freely.
Reduced the memory usage by 1/3 by doing manual allocations on the C++ side instead of passing a copy of the numpy array.

It also removes the screening based on atom distance, as the screening based on absolute overlap of the orbtials is much more efficient.
I will tag this PR with "removed functionality" as this might perhaps become relevant again.

### DIFF
--- a/docs/source/nitpick-exceptions
+++ b/docs/source/nitpick-exceptions
@@ -15,3 +15,4 @@ py:class numba.core.types.old_scalars.Integer
 py:class Function
 py:class pyscf.mp.mp2.MP2
 py:class RHF
+py:class quemb.molbe._cpp.eri_sparse_DF.SemiSparseSym3DTensor


src/quemb/molbe/eri_sparse_DF.py
```diff
@@ -1466,27 +594,22 @@ def transform_sparse_DF_integral_cpp(
...
-    cpp_transforms.set_log_level(logging.getLogger().getEffectiveLevel())
+    set_log_level(logging.getLogger().getEffectiveLevel())
     mol = mf.mol
     auxmol = make_auxmol(mf.mol, auxbasis=auxbasis)
 
     S_abs = approx_S_abs(mol)
     exch_reachable = _get_AO_per_AO(S_abs, AO_coeff_epsilon)
 
-    py_P_mu_nu = get_sparse_P_mu_nu(mol, auxmol, exch_reachable)
-    mu_nu_P = cpp_transforms.SemiSparseSym3DTensor(
```
mcocdawc: @mscho527 Here we had to duplicate memory allocation for (P | mu nu)

mscho527 (approved):
Looks good! Is it difficult to keep the tests in `tests_eri_sparse_DF` that are removed in this PR? i.e., testing those with cpp functions (maybe they are not exposed to python interface?) and setting it as tests that run only manually?

mcocdawc:
> Looks good! Is it difficult to keep the tests in `tests_eri_sparse_DF` that are removed in this PR? i.e., testing those with cpp functions (maybe they are not exposed to python interface?) and setting it as tests that run only manually?

Thank you! These tests tested functionality that I have not implemented in C++. Initially I did the screening based on atom distance and now make it based on the absolute overlap, which performs much better.
Hence, the atom distance screening is not even implemented in C++.

PR #239, opened 2025-10-22
P mu nu on the fly

mcocdawc:
Mayor change:

- Compute only the necessary (P | mu nu) tensors for every fragment on the fly. Reduces peak memory usage at the expense of computing time.

Minor change:

- In the computation for S_abs there is an additional renormalization step that accounts for the triangle inequality when using non cartesian basis sets.
- Declared some C++ stuff as `noexcept` and implemented `__repr__` magic methods for nicer strings.
- Additional cheap sanity checks in some C++ constructors.
- Removed some dead code.


mcocdawc:
> Cool! With this, perhaps in the near future, we might be able to deprecate `int-direct-DF`, right?

"int-direct-DF" is well tested and a reliable benchmark with sometimes better peak memory usage. I would keep it :-)

PR #252, opened 2025-12-02
Gradients clean

cariluo:
purpose of this draft: add a test for gradients

- [ ] require mypy again for executing pytest

Edit:
This branch records functionality related to force embedding, but there is no intention to ever merge it with main.


cariluo:
Note for the reader: we have determined that the original use of eigvecs to go from natural orbitals to embedding orbitals was correct, so I have reverted two previous commits.

cariluo:
Closing this PR because #272 accomplishes the same goal.

PR #259, opened 2026-03-12
Add ability to selectively initialize fragment

mscho527:
This adds an option to specify which fragment to perform ERI transformation when BE object is created. This functionality is useful for a number of projects going on in the group, i.e., gradients & ibc

For now, I am adding the functionality only for in-core and out-core-DF. Future PR should add them for other integral transformation routines as necessary.


cariluo (approved):
tried it myself, works fine

mscho527:
Note to myself: This PR is incomplete -- fragment initialization should also be modified to account for the lack of ERI elements for the non-selected fragments

PR #261, opened 2026-03-16
Fix bug with select init

mscho527:
#259 did not fully implement selective initialization, as it did not account for missing ERI in non-selected fragments. This PR remedies by skipping fragment HF for these non-selected fragments.

merge after #262

### DIFF
--- a/src/quemb/molbe/mbe.py
+++ b/src/quemb/molbe/mbe.py
@@ -1132,7 +1132,7 @@ class BE:
             assert_never(int_transform)
 
     @timer.timeit
-    def _initialize_fragments(self, file_eri: h5py.File, restart: bool):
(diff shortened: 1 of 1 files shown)


cariluo (approved):
yup, my only question is: does it still make sense to compute `ebe_hf` and then get the large `hf_err` for the case of not initializing all fragments? 
e.g. this is what I get for my test system 
```
all_frag_init is False
HF-in-HF error                 :  -1.6587e+02 Ha
```

mscho527:
I set the warning to not show up if not all fragments are initialized but kept the HF-in-HF error to show up for now because it doesn't require additional computational time. If we absolutely find no use in the future, it can be suppressed.

cariluo:
[like]  Carina Luo reacted to your message:
________________________________

PR #262, opened 2026-03-16
Update pandas intersphinx url

mscho527:
Fixes pandas intersphinx url to resolve docs build issue

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -67,7 +67,7 @@ templates_path = ["_templates"]
 
 intersphinx_mapping = {
     "numpy": ("https://numpy.org/doc/stable/", None),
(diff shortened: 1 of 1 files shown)


nwhel (approved):
It's a bit odd, as I am able to get the problematic inventory file from the old URL on my browser, but it looks like this URL works and makes sphinx happy so I don't see a reason not to change it.

mscho527:
> It's a bit odd, as I am able to get the problematic inventory file from the old URL on my browser, but it looks like this URL works and makes sphinx happy so I don't see a reason not to change it.

[https://pandas.pydata.org/pandas-docs/stable/object.inv](https://pandas.pydata.org/pandas-docs/stable/object.inv) does not contain intersphinx reference anymore. If you have DNS cache, it might still be pointing to the file.

PR #263, opened 2026-04-02
add periodic integral direct CCGDF

mscho527:
While implementing #260, I realized that there are additional considerations with pbc GDF, such as avoiding the G=0 divergence. As a stepping stone, I am adding integral-direct CCGDF (analogous to the int-direct-DF in molecular code) first. This initially targets gamma point to support sparse-DF implementation, but should be easier than sparse-DF to add k-point sampling.

Future PR should take some aspects back from the PBC code to molecular code, such as `eigh` fallback option if `cholesky` fails due to non positive definite metric `(P|Q)`.

### DIFF
--- a/src/quemb/kbe/eri_onthefly.py
+++ b/src/quemb/kbe/eri_onthefly.py
@@ -150,7 +150,7 @@ def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
             logger.debug("(μν|G) -> (ij|G) for frag #%d", fragidx)
             Gqi = ints @ Fobjs[fragidx].TA
             Giq = Gqi.transpose(0, 2, 1)
-            pqG_frag[fragidx][start:end, :, :] = Giq @ Fobjs[fragidx].TA
+            pqG_frag[fragidx][start:end, :, :] = Giq @ Fobjs[fragidx].TA.conj()
 
     end = 0
     blockranges = [
@@ -175,7 +175,7 @@ def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
             logger.debug("(μν|P) -> (ij|P) for frag #%d", fragidx)
             Lqi = ints @ Fobjs[fragidx].TA
             Liq = Lqi.transpose(0, 2, 1)
-            Lij = Liq @ Fobjs[fragidx].TA
+            Lij = Liq @ Fobjs[fragidx].TA.conj()
 
             # Add in contributions from the reciprocal space
             pqL_frag[fragidx][start:end, :, :] = Lij + (
--- a/src/quemb/kbe/eri_onthefly.py
+++ b/src/quemb/kbe/eri_onthefly.py
@@ -1,6 +1,6 @@
 import logging
 
-from numpy import complex128, sqrt, zeros
+from numpy import abs, complex128, sqrt, zeros
 from pyscf import lib
 from pyscf.ao2mo.addons import restore
 from pyscf.pbc.df.df import make_auxcell, make_modrho_basis
@@ -42,7 +42,7 @@ def _j2c_cholesky_or_eig(j2c):
             "decomposition."
         )
         d, V = eigh(j2c)  # V d V.T = (P|Q)
-        return V[:, d > 1e-14] / sqrt(d[d > 1e-14]) @ V[:, d > 1e-14].T, False
+        return V[:, d > 1e-14] / sqrt(d[d > 1e-14]) @ V[:, d > 1e-14].conj().T, False
 
 
 def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
@@ -163,7 +163,7 @@ def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
         for x, y in lib.prange(
             0,
             len(Gv),
-            block_step_size(len(Fobjs), len(Gv), mf.cell.nao, dtype=complex128),
+            block_step_size(len(Fobjs), len(Gv), mf.cell.nao, datatype=complex128),
         )
     ]
 
@@ -224,7 +224,7 @@ def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
             bb = j2c @ b
         logger.debug("Finished obtaining B_{ij}^{L} for frag #%d", fragidx)
         eri_nosym = bb.T @ bb
-        if (eri_nosym.imag > 1e-6).any():
+        if (abs(eri_nosym.imag) > 1e-6).any():
             raise ValueError(
                 f"Imaginary part of ERI is larger than 1e-6 for frag #{fragidx}."
             )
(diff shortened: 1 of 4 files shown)


mscho527:
@hannahHYL will tag along with @bhanscam for the review to learn how the ERI handling works!

src/quemb/molbe/eri_onthefly.py
```diff
@@ -15,6 +15,33 @@
...
+        Number of auxiliary basis functions
+    nao : int
+        Number of atomic orbitals
+    """
+    return max(
+        1,
+        int(
+            settings.INTEGRAL_TRANSFORM_MAX_MEMORY
+            * 1e9
+            / zeros((1), dtype=dtype).nbytes  # estimate size based on dtype
```
bhanscam: Instead of allocating memory to create an array, I think you can replace this with:
`np.dtype(dtype).itemsize`

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+        b = pqL_frag[fragidx].reshape(auxcell.nao, -1)
+        if ischol:
+            bb = solve_triangular(
+                j2c, b, lower=True, overwrite_b=True, check_finite=False
+            )
+        else:
+            bb = j2c @ b
+        logger.debug("Finished obtaining B_{ij}^{L} for frag #%d", fragidx)
+        eri_nosym = bb.T @ bb
+        if (eri_nosym.imag > 1e-6).any():
```
bhanscam: Is it possible to have a large negative imaginary value here?

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+        low = cholesky(j2c, lower=True)
+        logger.debug("(P|Q) positive definite. Cholesky decomposition successful.")
+        return low, True
+    except LinAlgError:
+        logger.debug(
+            "Cholesky decomposition of (P|Q) failed. Falling back to eigenvalue"
+            "decomposition."
+        )
+        d, V = eigh(j2c)  # V d V.T = (P|Q)
+        return V[:, d > 1e-14] / sqrt(d[d > 1e-14]) @ V[:, d > 1e-14].T, False
```
bhanscam: Probably not an issue in the current use but the rest of your added functions support complex datatypes so `V.conj().T` would be more consistent.

src/quemb/molbe/eri_onthefly.py
mscho527: good point, thanks!

src/quemb/kbe/eri_onthefly.py
mscho527: yes good catch!

src/quemb/kbe/eri_onthefly.py
mscho527: another good catch! thanks!

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+            mf.cell,
+            chgcell,
+            mf.cell._add_suffix("int3c2e"),
+            "s1",
+            shls_slice=shls_slice,
+        )  # Remove the part calculated in the reciprocal space to avoid double counting
+
+        logger.debug("Finish calculating (μν|P) for range %s", aux_range)
+
+        return ints.reshape(-1, mf.cell.nao, mf.cell.nao)  # TODO: ij pair to i, j
```
bhanscam: What ordering/shape does `aux_e2()` return? This reshape silently assumes `(naux, nao*nao)` but looking at the pyscf docs it is `(nao*nao, naux)` ordering. Either way I suggest adding an assertion with the expected size.

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+    pqL_frag = [
+        zeros((auxcell.nao, fragobj.nao, fragobj.nao), dtype=complex128)
+        for fragobj in Fobjs
+    ]  # place to store fragment (pq|L)
+
+    j2c = ccgdfbuilder.get_2c2e(
+        zeros((1, 3))
+    )  # (P|Q) accounting for periodic images and G=0 divergence
+    # TODO: reexamine kpoint, as (P|Q) depends on momentum transfer
+    j2c, ischol = _j2c_cholesky_or_eig(j2c[0])  # TODO: kpt
```
bhanscam, line 161: Not a big fan of having a variable meaning shift mid-function: `j2c` starts as the metric matrix then gets replaced with `L` after the cholesky decomp. I understand why you do that so if renaming isn't possible could you add a comment to clarify?

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+
+        return ints.reshape(-1, mf.cell.nao, mf.cell.nao)  # TODO: ij pair to i, j
+
+    def calculate_Gpq_pbcFS(pw_range):
+        r"""Internal function to calculate the 3-center integrals for a given range of
+        plane wave indices for periodic systems (in Fourier space)
+
+        Parameters
+        ----------
+        aux_range : tuple of int
```
bhanscam: argument is `pw_range` but docstring says `aux_range`

src/quemb/kbe/eri_onthefly.py
```diff
@@ -0,0 +1,234 @@
...
+    except LinAlgError:
+        logger.debug(
+            "Cholesky decomposition of (P|Q) failed. Falling back to eigenvalue"
+            "decomposition."
+        )
+        d, V = eigh(j2c)  # V d V.T = (P|Q)
+        return V[:, d > 1e-14] / sqrt(d[d > 1e-14]) @ V[:, d > 1e-14].conj().T, False
+
+
+def integral_direct_DF(mf, Fobjs, file_eri, auxbasis=None):
```
bhanscam, line 48: Since this is intended to the gamma-point with k-point sampling coming later, I suggest enforcing the currently supported conditions by raising a `NotImplementedError` if they aren't met.

bhanscam (approved):
Nice work! Math all looks good. Just a few consistency things to check/fix.

src/quemb/kbe/eri_onthefly.py
mscho527: You are right -- i am now surprised that this worked lol. I fixed and added assert

src/quemb/kbe/eri_onthefly.py
mscho527: Yeah this was because j2c has the leading memory scale of the routine (N_AO^2). Python garbage collection is.... great.... so I renamed so that we don't waste more memory than we need. I added the info as comment.

src/quemb/kbe/eri_onthefly.py
mscho527: Fixed. Good catch!

src/quemb/kbe/eri_onthefly.py
mscho527: Good call. I added the exception at the initialization phase of kbe.py instead, so that the conditions are enforced before we burn trees for no reason.

bhanscam: The trees thank you :)

hannahHYL (approved):

PR #265, opened 2026-04-13
Numerical Jacobian for density matching

### LINKED ISSUE #34: QN Trust Region Broyden with Numerical Initial Jacobian
Discussed in subgroups many months ago that we could try adding numerical initial jacobian for the trust-region Broyden optimizer.

Very low priority, but perhaps worth trying at some point

mscho527:
This PR adds on to #250 and allows numerical Jacobian routine for density matching. While this is much slower than existing analytical options, it is helpful for problems that are very difficult to converge.

Subsequent PR will replace the for loop in compute_numerical_jacobian::91 to a parallel dispatch.

Resolves #34

### DIFF
--- a/src/quemb/molbe/be_parallel.py
+++ b/src/quemb/molbe/be_parallel.py
@@ -297,10 +297,14 @@ def run_solver(
             use_cumulant,
             eri_file,
         )
-    if eeval and not ret_vec:
-        return e_f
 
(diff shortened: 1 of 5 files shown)


hannahHYL (approved):
Thanks for the slack and blackboard explanation :) I don't have any additional comments.

PR #266, opened 2026-04-16
quick fix to enable selective initialization for int-direct-DF

mscho527:
necessary for gradient work

### DIFF
--- a/src/quemb/molbe/mbe.py
+++ b/src/quemb/molbe/mbe.py
@@ -300,6 +300,7 @@ class BE:
         if initialize_fragment_idx is not None and int_transform not in [
             "in-core",
             "out-core-DF",
+            "int-direct-DF",
         ]:
             raise NotImplementedError(
                 "Selective fragment initialization is only implemented for 'in-core'"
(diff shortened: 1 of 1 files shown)


src/quemb/molbe/mbe.py
```diff
@@ -302,8 +302,8 @@ def __init__(
             "out-core-DF",
         ]:
             raise NotImplementedError(
-                "Selective fragment initialization is only implemented for 'in-core' "
-                "and 'out-core-DF' integral transformations."
+                "Selective fragment initialization is only implemented for 'in-core'"
+                ", 'out-core-DF', and 'int-direct-DF' integral transformations."
```
cariluo, line 307: this line says there are 3 options for the integral transform but line 300 checks 2 options for integral transform, this leads to the error message printing when I try to do `int-direct-DF`

cariluo (changes requested):

src/quemb/molbe/mbe.py
mscho527: oh yes what was i thinking -- just fixed it

cariluo (approved):

PR #269, opened 2026-06-01
simple numerical geometry optimization framework

bhanscam:
This is the start of building a geometry scanner/optimizer using numerical gradients from BE. Initially this builds the template for feeding a custom energy function into the pyscf finite_diff module and using the result to compute gradients, hessians and run geometry scans and optimizations through the pyscf geometry optimization modules. Future PRs will implement BE as the custom energy function.

Desired functionality:
- numerical gradients of a custom energy function using finite_diff pyscf module
- scanner functionality for energy and gradient point calculations
- integration with pyscf geometry optimization modules
- control over coordinate displacement magnitude
- information sharing between finite_diff module and custom energy function

### DIFF
--- a/example/molbe_h2_scanner.py
+++ b/example/molbe_h2_scanner.py
@@ -1,11 +1,12 @@
 # Illustrates nuclear gradient and hessian computations as well as geometry optimizaton
 # using a custom energy function fed through pyscf's finite difference driver
+# Author(s): Beck Hanscam
 
 import numpy as np
 from pyscf import grad, gto, scf
 from pyscf.tools import finite_diff
 
-from quemb.molbe.geomopt import Energy, energy_hf
+from quemb.molbe.scanner import Energy, energy_hf
 
 # =========================
 #  Gradients
--- a/src/quemb/molbe/geomopt.py
+++ b/src/quemb/molbe/geomopt.py
@@ -1,8 +1,9 @@
 import inspect
 import os
+from dataclasses import dataclass, field
 
 import numpy as np
-from pyscf import lib, scf
+from pyscf import gto, lib, scf
 
 
 def energy_hf(mol, fd_info=None):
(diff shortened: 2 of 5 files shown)


src/quemb/molbe/scanner.py
mscho527, line 1: nitpick, but i think something like "scanner.py" or "energy_scanner.py" fits better as the module name, as this alone doesn't (and we don't plan to) perform geometry optimization.

src/quemb/molbe/scanner.py
```diff
@@ -0,0 +1,212 @@
+import inspect
+import os
+
+import numpy as np
+from pyscf import lib, scf
+
+
+def energy_hf(mol, fd_info=None):
```
mscho527, line 11: Instead of a dictionary, can we use argument class? This ensures that all arguments are passed along (or set to default values). See ChemGenArgs for reference.

src/quemb/molbe/geomopt.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Returns
+    ------
+    float
+        Converged RHF total energy in Hartree
+    """
+    if fd_info is not None:
+        if not np.allclose(
+            fd_info["ref_coords"],
+            fd_info["ref_mol"].atom_coords(),
```
mscho527: do we need ref_coords? or can we let the argument class (that i suggested above) automatically set this to ref_mol.atom_coords()?

example/molbe_h2_scanner.py
mscho527, line 1: Can you add a test, perhaps based on this example script?

src/quemb/molbe/scanner.py
bhanscam: Oh that's better, I agree!

src/quemb/molbe/geomopt.py
bhanscam: No we don't need ref_coords if we have ref_mol. I just threw it in there for now as an easy sanity check.

example/molbe_h2_scanner.py
bhanscam: Should I build a test for this version or wait until we hook up BE? Or have tests for both?

bhanscam: Added a test using H2 for the gradient, hessian, energy scanner and gradient scanner. I can add a separate test using H8 when we add a BE energy function.

mscho527 (approved):

cariluo (approved):

PR #270, opened 2026-06-02
Fix atol for failing localization tests to reflect BE optimization tolerance.

mscho527:
There are some failing tests in the localization tests. Superficially, this comes from numerical error coming from new numpy version. The failing tests compare the optimized BE correlation energy, which are only converged to 1e-6. Tests were applied with atol = 1e-8 and rtol = 0, and therefore failed due to an order 1e-6 error. We should apply tolerance that we can expect from the BE optimization routine, rather than a overly strict criteria.

### DIFF
--- a/tests/loc_test.py
+++ b/tests/loc_test.py
@@ -1,6 +1,5 @@
 # Tests the standard localization and IAO routines for hexene
 
-import os
 import unittest
 from typing import Literal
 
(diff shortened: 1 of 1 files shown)


tests/loc_test.py
hannahHYL: Does this threshold need to be adjusted as well?

mscho527: ah good point! adjusted this thanks!

mscho527:
I think we could keep the thresholds for other tests as is for now, since we just want to resolve the failing test. We might need to, in the future, if issues arise.

PR #271, opened 2026-06-04
add BE energy to numerical gradient framework

bhanscam:
This is the second stage of building a geometry scanner, adding a BE energy function and interface to the framework established in #269. 

Additional BE-specific consideration is user control of BE arguments and keywords outside of the scanner framework, including inputs to `fragmentate`, `oneshot`, and `optimize`.

### DIFF
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -54,7 +54,7 @@ ref_grad_corr = mc_grad.kernel()
 # numerical gradients of mol0
 # BEn-CCSD oneshot, n=1,2,3
 #
-print("\nnumerical CCSD gradient for mol0:")
+print("\nnumerical BE-CCSD gradient for mol0:")
 rms = []
 for n_BE in [1, 2, 3]:
     be_args = BEArgs(
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -64,7 +64,9 @@ for n_BE in [1, 2, 3]:
         optimize=False,
         additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
     )
-    energy_method = Energy(mol0, energy_be, be_args=be_args, ref_data_func=be_ref_data)
+    energy_method = Energy(
+        mol0, energy_be, energy_args=be_args, ref_data_func=be_ref_data
+    )
     grad_method = finite_diff.Gradients(energy_method)
     grad_method.displacement = 1e-4  # Gradients class default is 1e-2 Bohr
     fd_grad_fromObj = grad_method.kernel()
@@ -90,7 +92,7 @@ be_args = BEArgs(
     only_chem=True,
     additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
 )
-energy_method = Energy(mol0, energy_be, be_args=be_args, ref_data_func=be_ref_data)
+energy_method = Energy(mol0, energy_be, energy_args=be_args, ref_data_func=be_ref_data)
 energy_scanner = energy_method.as_scanner()
 es_E0 = energy_scanner(mol0)
 es_E1 = energy_scanner(mol1)
@@ -108,7 +110,7 @@ be_args = BEArgs(
     optimize=False,
     additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
 )
-energy_method = Energy(mol0, energy_be, be_args=be_args, ref_data_func=be_ref_data)
+energy_method = Energy(mol0, energy_be, energy_args=be_args, ref_data_func=be_ref_data)
 grad_method = finite_diff.Gradients(energy_method)
 grad_method.displacement = 1e-4  # Gradients class default is 1e-2 Bohr
 grad_scanner = grad_method.as_scanner()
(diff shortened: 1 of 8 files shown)


example/molbe_h6_scanner.py
```diff
@@ -0,0 +1,136 @@
...
+mc = cc.CCSD(mf)
+mc.kernel()
+mc_grad = mc.nuc_grad_method()
+ref_grad_corr = mc_grad.kernel()
+
+#
+# numerical gradients of mol0
+# BEn-CCSD oneshot, n=1,2,3
+#
+print("\nnumerical CCSD gradient for mol0:")
```
cariluo: Could we call this "numerical BE CCSD gradient for mol0"? To me, numerical CCSD gradient is what was included in #269

cariluo (commented):
And to check my understanding - the test only does oneshot, but in principle, a test could be added for the chemical potential matching analogue in a straightforward way since you've set it up nicely?

example/molbe_h6_scanner.py
bhanscam: Yep yep

bhanscam:
> And to check my understanding - the test only does oneshot, but in principle, a test could be added for the chemical potential matching analogue in a straightforward way since you've set it up nicely?

Yes absolutely. I didn't include matching in the test because it takes a while to run. I can add a matching test and just keep it turned off for now.

src/quemb/molbe/scanner.py
```diff
@@ -34,6 +37,149 @@ def energy_hf(mol, fd_info=None):
...
+    # additional optimize keywords
+    optimize: bool = True
+    only_chem: bool = False
+    method: str = "QN"
+    conv_tol: float = 1.0e-6
+    relax_density: bool = False
+    jac_solver: str = "HF"
+    max_iter: int = 500
+    trust_region: bool = False
+    step_size: float = 1e-6
```
mscho527: I need to think a little bit more about this, but if we want a BEArgs dataclass, I think this should live somewhere near where BE arguments are defined (ex., mbe.py) for code maintainability. If someone adds (or removes) an argument from mbe, we want the BEArgs class to be updated simultaneously.

src/quemb/molbe/scanner.py
```diff
@@ -60,7 +207,9 @@ class Energy(lib.StreamObject):
     difference gradient or Hessian drivers.
     """
 
-    def __init__(self, mol, energy_func, displacement=1e-4):
+    def __init__(
+        self, mol, energy_func, displacement=1e-4, ref_data_func=None, **energy_kwargs
```
mscho527: Unless we can't avoid it, we should stay away from using `**energy_kwargs`-like structure. This makes it more difficult to identify which arguments are allowed for the function and are actually used. I think you can just explicitly put BEArgs and FDInfo, right?

tests/molbe_h6_scanner_test.py
```diff
@@ -0,0 +1,168 @@
...
+    # BE3-CCSD oneshot numerical gradient from scanner
+    grad_scanner = grad_method.as_scanner()
+    E1, fd_grad1 = grad_scanner(mol1)
+    assert np.isclose(E1, -3.23567708251885)
+    assert np.isclose(
+        np.sqrt(np.mean((ref_grad_corr1 - fd_grad1) ** 2)), 5.8664061568865036e-08
+    )
+
+
+def test_numerical_be_gradient_withmatching():
```
mscho527, line 100: This one takes quite a bit (see GHA pytest logs; adds 40s to < 8m total test), and from what I understand, has most of its functionality covered by the test without matching and other BE optimization tests outside this file. I suggest adding the expensive test tag, so that it is triggered only on pr merge.

mscho527 (changes requested):

src/quemb/molbe/scanner.py
bhanscam: A dataclass was the only way I could come up with passing all the necessary info through the finite difference driver with the functionality we need. Happy to move it to mbe (and rename it so it's clearly for the scanner and not general use) or handle the user arguments in a completely different way if you have a better idea.

mscho527: Yeah, agreed that dataclass is a good solution here! I support moving it to mbe. Let's keep it for scanner only for now, but we might even want to use it for general use later.

mscho527: Oh and to clarify my comment, I think the name of the data lclass (BEArgs) is good -- probably doesn't need to be changed. I was mostly worried where it should live.

src/quemb/molbe/scanner.py
bhanscam: That structure allows the `Energy` class to be fully flexible to the necessary user inputs for whatever energy function is passed through it, keeping this setup functional for arbitrary energy functions. For example `energy_hf()` doesn't need BEArgs or any other user controls. Instead of explicitly passing BEArgs, I could enforce the energy function to always take in a dataclass argument which can be empty when not needed. Is that a better setup since this is meant mostly for BE-specific use?

mscho527: I think I prefer
```
def __init__(
..., be_args = None, fd_info = None
)
```
to blanket **kwargs, just to make it clear to users that those are the available argument(s) to the function.

tests/molbe_h6_scanner_test.py
bhanscam: That is a typo haha. I didn't mean to allow that test to run at all since its functionality is covered by other tests. I only wrote it out for completeness (see previous conversation with Carina). Should I add the expensive test tag or just not allow it to run? Or both?

mscho527: Expensive tag! I don't think it makes sense to add a test and not allow it to run.

mscho527 (approved):
Cool, thanks for the changes!

Issue #208, opened 2025-08-19
`mypy` errors for `graphfrag.py`


mscho527:
There are few `mypy` errors that show up for `graphfrag.py`. It looks like it is coming from the fact that line 209 creates list of tuples, whereas `draw_networkx_nodes` expect `Mapping` (as defined in `nx_pylab.pyi`). Line 217 actually creates `Mapping`.

Possible fix would be to modify line 209 so that it creates appropriate `Mapping` and reactivate the ignored errors (copied below).

```
src/quemb/molbe/graphfrag.py:226: error: Argument 2 to "draw_networkx_nodes" has incompatible type "list[tuple[Any, Any]]"; expected "Mapping[Any, Collection[float]]"  [arg-type]
src/quemb/molbe/graphfrag.py:235: error: Argument 2 to "draw_networkx_nodes" has incompatible type "list[tuple[Any, Any]]"; expected "Mapping[Any, Collection[float]]"  [arg-type]
src/quemb/molbe/graphfrag.py:242: error: No overload variant of "draw_networkx_edges" matches argument types "Graph[Any]", "list[tuple[Any, Any]]", "bool", "Sequence[Any]", "int", "float", "tuple[float, float, float]", "str"  [call-overload]
src/quemb/molbe/graphfrag.py:242: note: Possible overload variants:
src/quemb/molbe/graphfrag.py:242: note:     def [_Node: Hashable] draw_networkx_edges(G: Graph[_Node], pos: Mapping[_Node, Collection[float]], edgelist: Collection[_Node | Hashable] | None = ..., width: float | Collection[float] = ..., edge_color: tuple[float, float, float] | str | str | tuple[float, float, float, float] | tuple[tuple[float, float, float] | str, float] | tuple[tuple[float, float, float, float], float] | Collection[tuple[float, float, float] | str | str | tuple[float, float, float, float] | tuple[tuple[float, float, float] | str, float] | tuple[tuple[float, float, float, float], float]] = ..., style: str | Collection[str] = ..., alpha: float | Collection[float] | None = ..., arrowstyle: str | Collection[str] | None = ..., arrowsize: float | list[int] | list[float] = ..., edge_cmap: Colormap | None = ..., edge_vmin: float | None = ..., edge_vmax: float | None = ..., ax: Axes | None = ..., arrows: None = ..., label: str | None = ..., node_size: float | Collection[float] = ..., nodelist: Sequence
[...]

PR #272, opened 2026-06-09
add force embedding to numerical gradient framework 

cariluo:
This is the third stage of building a geometry scanner, adding a force embedding fragment energy function to the framework established in #269 and #271. 

Additional force embedding-specific consideration is storage of the reference information (mol, fobj, Fobjs, frag_per_atom) so that it does not need to recomputed and user selection of the orbital alignment procedure.

### DIFF
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -91,6 +91,7 @@ for n_BE in [1, 2, 3]:
         use_cumulant=True,
         optimize=False,
         additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
+        orbital_alignment="block-diagonal",
     )
     energy_method = Energy(
         mol0, energy_force_emb, energy_args=be_args, ref_data_func=force_emb_ref_data
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -7,7 +7,14 @@ from pyscf import cc, gto, scf
 from pyscf.tools import finite_diff
 
 from quemb.molbe.chemfrag import ChemGenArgs
-from quemb.molbe.scanner import BEArgs, Energy, be_ref_data, energy_be
+from quemb.molbe.scanner import (
+    BEArgs,
+    Energy,
+    be_ref_data,
+    energy_be,
+    energy_force_emb,
+    force_emb_ref_data,
+)
 
 # build several different geometries
 rng = np.random.default_rng()
@@ -55,7 +62,7 @@ ref_grad_corr = mc_grad.kernel()
 # BEn-CCSD oneshot, n=1,2,3
 #
 print("\nnumerical BE-CCSD gradient for mol0:")
-rms = []
+rms_be = []
 for n_BE in [1, 2, 3]:
     be_args = BEArgs(
         n_BE=n_BE,
@@ -70,7 +77,28 @@ for n_BE in [1, 2, 3]:
     grad_method = finite_diff.Gradients(energy_method)
     grad_method.displacement = 1e-4  # Gradients class default is 1e-2 Bohr
     fd_grad_fromObj = grad_method.kernel()
-    rms.append(np.sqrt(np.mean((ref_grad_corr - fd_grad_fromObj) ** 2)))
+    rms_be.append(np.sqrt(np.mean((ref_grad_corr - fd_grad_fromObj) ** 2)))
+
+# numerical gradients of mol0
+# BEn-CCSD force embedding, n=1,2,3
+#
+print("\nnumerical BE-CCSD force embedding gradient for mol0:")
+rms_fe = []
+for n_BE in [1, 2, 3]:
+    be_args = BEArgs(
+        n_BE=n_BE,
+        solver="CCSD",
+        use_cumulant=True,
+        optimize=False,
+        additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
+    )
+    energy_method = Energy(
+        mol0, energy_force_emb, energy_args=be_args, ref_data_func=force_emb_ref_data
+    )
+    grad_method = finite_diff.Gradients(energy_method)
+    grad_method.displacement = 1e-4
+    fd_grad_fromObj = grad_method.kernel()
+    rms_fe.append(np.sqrt(np.mean((ref_grad_corr - fd_grad_fromObj) ** 2)))
 
 
 # =========================
@@ -124,7 +152,17 @@ gs_E2, gs_grad2 = grad_scanner(mol2)
 
 print("\n\n___NUMERICAL GRADIENTS___")
 for n_BE in [1, 2, 3]:
-    print(f"BE{n_BE}-CCSD RMSE from analytic benchmark for mol0: {rms[n_BE - 1]:.12e}")
+    print(
+        f"BE{n_BE}-CCSD RMSE from analytic benchmark for mol0: {rms_be[n_BE - 1]:.12e}"
+    )
+
+print()
+
+for n_BE in [1, 2, 3]:
+    print(
+        f"BE{n_BE}-CCSD force embedding RMSE"
+        f"\nfrom analytic benchmark for mol0: {rms_fe[n_BE - 1]:.12e}"
+    )
 
 print("\n___ENERGY SCANNER___")
 print(f"BE2-CCSD (with chempot matching) energy of mol0: {es_E0:.8f}")
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -10,7 +10,6 @@ from quemb.molbe.chemfrag import ChemGenArgs
 from quemb.molbe.mbe import BEArgs
 from quemb.molbe.scanner import (
     Energy,
-    ForceEmbArgs,
     be_ref_data,
     energy_be,
     energy_force_emb,
@@ -86,13 +85,12 @@ for n_BE in [1, 2, 3]:
 print("\nnumerical BE-CCSD force embedding gradient for mol0:")
 rms_fe = []
 for n_BE in [1, 2, 3]:
-    force_emb_args = ForceEmbArgs(
+    force_emb_args = BEArgs(
         n_BE=n_BE,
         solver="CCSD",
         use_cumulant=True,
         optimize=False,
         additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
-        orbital_alignment="block-diagonal",
     )
     energy_method = Energy(
         mol0,
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -10,10 +10,10 @@ from quemb.molbe.chemfrag import ChemGenArgs
 from quemb.molbe.mbe import BEArgs
 from quemb.molbe.scanner import (
     Energy,
+    be_frag_ref_data,
     be_ref_data,
     energy_be,
-    energy_force_emb,
-    force_emb_ref_data,
+    energy_be_frag,
 )
 
 # build several different geometries
@@ -94,9 +94,9 @@ for n_BE in [1, 2, 3]:
     )
     energy_method = Energy(
         mol0,
-        energy_force_emb,
+        energy_be_frag,
         energy_args=force_emb_args,
-        ref_data_func=force_emb_ref_data,
+        ref_data_func=be_frag_ref_data,
     )
     grad_method = finite_diff.Gradients(energy_method)
     grad_method.displacement = 1e-4
--- a/example/molbe_h6_scanner.py
+++ b/example/molbe_h6_scanner.py
@@ -57,6 +57,15 @@ mc.kernel()
 mc_grad = mc.nuc_grad_method()
 ref_grad_corr = mc_grad.kernel()
 
+mf = scf.RHF(mol1)
+mf.conv_tol = 1e-12
+mf.kernel()
+mc = cc.CCSD(mf)
+mc = cc.CCSD(mf)
+mc.kernel()
+mc_grad = mc.nuc_grad_method()
+ref_grad_corr1 = mc_grad.kernel()
+
 #
 # numerical gradients of mol0
 # BEn-CCSD oneshot, n=1,2,3
@@ -148,6 +157,27 @@ grad_scanner = grad_method.as_scanner()
 gs_E1, gs_grad1 = grad_scanner(mol1)
 gs_E2, gs_grad2 = grad_scanner(mol2)
 
+#
+# gradient point calculations using as_scanner()
+# BE3-CCSD force embedding
+#
+print("\nUsing grad_method.as_scanner():")
+be_args = BEArgs(
+    n_BE=3,
+    solver="CCSD",
+    use_cumulant=True,
+    optimize=False,
+    additional_args=ChemGenArgs(h_treatment="treat_H_like_heavy_atom"),
+)
+energy_method = Energy(
+    mol0, energy_be_frag, energy_args=be_args, ref_data_func=be_frag_ref_data
+)
+grad_method = finite_diff.Gradients(energy_method)
+grad_method.displacement = 1e-4  # Gradients class default is 1e-2 Bohr
+grad_scanner = grad_method.as_scanner()
+fe_gs_E1, fe_gs_grad1 = grad_scanner(mol1)
+fe_gs_E2, fe_gs_grad2 = grad_scanner(mol2)
+
 
 # =========================
 #  Print results
@@ -174,6 +204,13 @@ print(f"BE2-CCSD (with chempot matching) energy of mol2: {es_E2:.8f}")
 
 print("\n___GRADIENT SCANNER___")
 print(f"BE3-CCSD energy of mol1: {gs_E1:.8f}")
-print("BE3-CCSD numerical grad of mol1: \n", gs_grad1)
-print(f"\nBE3-CCSD energy of mol2: {gs_E2:.8f}")
+print(f"BE3-CCSD energy of mol2: {gs_E2:.8f}")
+
+print(f"\nBE3-CCSD force embedding energy of mol1: {fe_gs_E1:.8f}")
+print(f"BE3-CCSD force embedding energy of mol2: {fe_gs_E2:.8f}")
+
+print("\nBE3-CCSD numerical grad of mol1: \n", gs_grad1)
 print("BE3-CCSD numerical grad of mol2: \n", gs_grad2)
+
+print("\nBE3-CCSD force embedding numerical grad of mol1: \n", fe_gs_grad1)
+print("BE3-CCSD force embedding numerical grad of mol2: \n", fe_gs_grad2)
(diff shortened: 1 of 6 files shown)


src/quemb/molbe/mbe.py
```diff
@@ -145,6 +148,9 @@ class BEArgs:
     trust_region: bool = False
     step_size: float = 1e-6
 
+    # force embedding keywords
```
mscho527: This is an option that only applies for the gradients, so I think it should live in scanner.py, not mbe.py. We have these options here so that any changes in mbe.py can easily be reflected here. Likewise, any argument that relates to scanner.py should live there.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
     return mybe.ebe_tot
 
 
+def energy_force_emb(mol, energy_args=None, fd_info=None):
```
mscho527: I think we determined that we should only implement one orbital alignment option for the production code. If the plan changed, we should explicitly put the alignment option as a required input in this function, rather than something that gets passed as part of energy_args. Also, we should test every option.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+    if energy_args.orbital_alignment is None:
+        pass  # Do not do orbital alignment
+
+    elif energy_args.orbital_alignment == "block-diagonal":
+        # Get the natural orbitals in terms of embedding orbitals
+        # for the reference geometry
+        C_ = multi_dot(
+            (ref_mybe.Fobjs[frag_idx].TA.T, ref_mybe.S, ref_mybe.C[:, : ref_mybe.Nocc])
+        )
+        P_ = C_ @ C_.T
```
mscho527: As discussed before, this is just dm0, so we can avoid recomputing

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+        # and the AO AO overlap matrix between the two geometries
+        TA_ref = ref_mybe.Fobjs[frag_idx].TA @ eigvecs
+        nsocc = ref_mybe.Fobjs[frag_idx].nsocc
+        S_cross = gto.intor_cross("int1e_ovlp", mol, fd_info.ref_mol)
+
+        # Solve the orthogonal procrustes problem
+        # for the occupied and virtual separately
+        H = mybe.C.T @ S_cross @ TA_ref
+        H[: mybe.Nocc, nsocc:] = 0
+        H[mybe.Nocc :, :nsocc] = 0
```
mscho527: Perhaps use a more descriptive variable name than `H`

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+            ref_mybe = fd_info.ref_data["ref_mybe"]
+            frag_per_atom = fd_info.ref_data["frag_per_atom"]
+        except KeyError as exc:
+            raise RuntimeError("missing reference BE info.") from exc
+
+    mf = scf.RHF(mol)
+    mf.verbose = 0
+    mf.kernel()
+
+    frag_idx = frag_per_atom[fd_info.atom_idx[0]]
```
mscho527, line 283: Shouldn't we check to make sure that the `fd_info.atom_idx` is all identical before pulling the first element?

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+
+    else:
+        raise ValueError(
+            "orbital_alignment must be None, 'block-diagonal', or 'basis-projection', "
+            f"got {energy_args.orbital_alignment!r}"
+        )
+
+    # Redo the integral transform and fragment initialization
+    # with the new TA
+    file_eri = h5py.File(mybe.eri_file, "w")
```
mscho527: We should put new integrals to a separate file and make sure that it is cleaned up after calculation to avoid scratch fill up. Perhaps use Oskar's scratch handling routine by creating a subdirectory.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+    mybe._eri_transform(energy_args.int_transform, mf._eri, file_eri, [frag_idx])
```
mscho527: We should probably batch (or likely handle all at once) ERI transformation so that we don't end up evaluating AO ERIs or DF three tensors O(3N) times.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+
+    mybe._initialize_fragments(file_eri, False, [frag_idx])
+    file_eri.close()
+
+    # Get the correlation energy for the fragment
+    eri = get_eri(fobj.dname, fobj.nao, eri_file=fobj.eri_file)
+    fobj._mf = get_scfObj(fobj.fock, eri, fobj.nsocc, dm0=fobj.dm0.copy())
```
mscho527: Okay for now, but might require `heff` if we work with chem pot

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+    # Redo the integral transform and fragment initialization
+    # with the new TA
+    file_eri = h5py.File(mybe.eri_file, "w")
+    mybe._eri_transform(energy_args.int_transform, mf._eri, file_eri, [frag_idx])
+
+    mybe._initialize_fragments(file_eri, False, [frag_idx])
+    file_eri.close()
+
+    # Get the correlation energy for the fragment
+    eri = get_eri(fobj.dname, fobj.nao, eri_file=fobj.eri_file)
```
mscho527: With ERIs on a separate file, `dname` should be modified appropriately

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +214,127 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+    mc.incore_complete = True
+
+    eri_embmo = mc.ao2mo()
+    eri_embmo.mo_energy = fobj._mf.mo_energy
+    eri_embmo.fock = np.diag(fobj._mf.mo_energy)
+
+    mc.kernel(eris=eri_embmo)
+    e_corr = mc.e_tot - fobj._mf.e_tot
+
+    return mf.e_tot + e_corr
```
mscho527: mc.e_tot + mf.energy_nuc()?

tests/molbe_h6_scanner_test.py
```diff
@@ -93,6 +100,57 @@ def test_numerical_be_gradient():
...
+    )
+
+    # CCSD analytic gradient of mol0
+    mf = scf.RHF(mol0)
+    mf.conv_tol = 1e-12
+    mf.kernel()
+    mc = cc.CCSD(mf)
+    mc.kernel()
+    mc_grad = mc.nuc_grad_method()
+    ref_grad_corr = mc_grad.kernel()
```
mscho527: This can probably be factored out into a new function since it is just repeated from other test.

mscho527 (dismissed):

src/quemb/molbe/scanner.py
bhanscam: I think it makes the most sense to only implement the basis projection version only and leave the other options in the gradients-clean branch for reference.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +213,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
     return mybe.ebe_tot
 
 
+def energy_force_emb(mol, energy_args=None, fd_info=None):
```
bhanscam: Can we name this differently to be consistent with the naming in this file? Maybe something like `energy_be_frag`? Something that indicates it is only using the BE fragment energy for the displaced geometry and not the full system like `energy_be` does.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +213,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+        Converged BE fragment energy in Hartree
+    """
+    if energy_args is None:
+        energy_args = BEArgs()
+
+    if fd_info is None:
+        raise RuntimeError("missing finite difference displacement info.")
+    else:
+        if fd_info.ref_mol is None:
+            raise RuntimeError("missing finite difference reference geometry.")
```
bhanscam, line 247: This function currently cannot handle multiple atoms being displaced at once, which is what occurs in a Hessian call. In that instance `fd_info.atom_idx` would contain more than one atom index. Add a check to ensure `fd_info.kind != multi_displacement`. Something like this
```
if fd_info.kind == "multi_displacement":
     raise RuntimeError(
         "energy_force_emb currently supports only single finite-difference "
         f"displacements; got {fd_info.kind!r}."
     )
```

src/quemb/molbe/scanner.py
bhanscam: Agreed. This function should exclusively be seeing `fd_info.atom_idx` as a one element list, anything else should return an error. If `fd_info.atom_idx=[]` then it is a reference geometry or scanner point call that was mishandled, and if `fd_info.atom_idx` has more than one element then it was a multi-displacement likely from a Hessian call or mishandled geometry optimization.

src/quemb/molbe/scanner.py
```diff
@@ -71,6 +83,65 @@ def be_ref_data(mol, energy_args=None):
...
+
+    fragmented = Fragmented.from_mole(
+        mol, n_BE=energy_args.n_BE, h_treatment=energy_args.additional_args.h_treatment
+    )
+    frag_per_atom = fragmented.get_frag_per_atom()
+
+    ref_dict = {
+        "ref_mol": mol,
+        "ref_fobj": ref_fobj,
+        "ref_mybe": ref_mybe,
```
bhanscam, line 131: The docstring says this returns "ref_Fobjs" and not "ref_mybe".

src/quemb/molbe/scanner.py
```diff
@@ -71,6 +83,65 @@ def be_ref_data(mol, energy_args=None):
...
+        initialize_fragment_idx=energy_args.initialize_fragment_idx,
+    )
+
+    fragmented = Fragmented.from_mole(
+        mol, n_BE=energy_args.n_BE, h_treatment=energy_args.additional_args.h_treatment
+    )
+    frag_per_atom = fragmented.get_frag_per_atom()
+
+    ref_dict = {
+        "ref_mol": mol,
```
bhanscam: To be safe, make this `mol.copy()`

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +221,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
     return mybe.ebe_tot
 
 
+def energy_force_emb(mol, energy_args=None, fd_info=None):
```
bhanscam: Can we name this differently to be consistent with the naming this file? Maybe something like `energy_be_frag`? (and similar for the ref_data function)

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +221,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+        auxbasis=energy_args.auxbasis,
+        nproc=energy_args.nproc,
+        ompnum=energy_args.ompnum,
+        initialize_fragment_idx=[frag_idx],
+    )
+
+    fobj = mybe.Fobjs[frag_idx]
+
+    if energy_args.orbital_alignment is None:
+        pass  # Do not do orbital alignment
```
bhanscam: Do we need the option to skip orbital alignment or can that functionality stay in the gradients-clean branch for now?

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +221,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+        # for the occupied and virtual separately
+        H = mybe.C.T @ S_cross @ TA_ref
+        H[: mybe.Nocc, nsocc:] = 0
+        H[mybe.Nocc :, :nsocc] = 0
+
+        U, _, Vt = svd(H, full_matrices=False)
+        R = U @ Vt
+
+        fobj.TA = mybe.C @ R @ eigvecs.T
+
```
bhanscam: See previous comment about not including the orbital alignment options in this PR

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +221,122 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+        )
+
+    # Redo the integral transform and fragment initialization
+    # with the new TA
+    file_eri = h5py.File(mybe.eri_file, "w")
+    mybe._eri_transform(energy_args.int_transform, mf._eri, file_eri, [frag_idx])
+
+    mybe._initialize_fragments(file_eri, False, [frag_idx])
+    file_eri.close()
+
```
bhanscam: You possibly have a stale `fobj` here? It is set before re-initializing and is then changed in place (right?) so I think it is okay, but consider adding a line after re-initialization to be safe:
`fobj = mybe.Fobjs[frag_idx]`

example/molbe_h6_scanner.py
bhanscam, line 1: The force embedding option should be able to do the gradient scanner functionality as well, but cannot do the energy scanner. Add a gradient scanner example.

tests/molbe_h6_scanner_test.py
bhanscam, line 1: Add a test for the gradient scanner functionality.

bhanscam (changes requested):

src/quemb/molbe/scanner.py
cariluo: Note from irl discussion: check for multi_disp as above, then assert that atom_idx has length 1

src/quemb/molbe/scanner.py
cariluo: Note from irl conversation on Thursday 6/11/26 evening: was told to disregard this comment for now

example/molbe_h6_scanner.py
cariluo: I tried to add a gradient scanner example, but the gradient scanner returns both an energy and a gradient, and I'm not sure that force embedding has an energy associated with the gradient.

```
gs_E2, gs_grad2 = grad_scanner(mol2)
```

I get an error at the assert because fd_info.atom_idx is empty.

```
 AssertionError: Expected fd_info.atom_idx to have length 1, but got 0: []                                                  
```

tests/molbe_h6_scanner_test.py
cariluo: See the comment on the above request for a gradient scanner example for the force embedding option.

cariluo:
I have now addressed or replied to all requested changes. The larger changes are:

- removed force embedding keywords from scanner.py and implemented only basis set projection orbital alignment option 
- checked that `fd_info.atom_idx` is a one element list 
- put new ERIs on a separate file and clean up after calculation, also modified `dname`

Unaddressed difficulties:
- The ability to do force embedding with the gradient scanner functionality (see reply above)
- I get a mypy error on `graphfrag.py` (not sure how to handle this one, as this PR is not about graphfrag)

example/molbe_h6_scanner.py
bhanscam: The energy coming out of the gradient scanner won't follow the same setup as `energy_hf` or `energy_be` but you can still return the HF energy. Add a condition in `energy_be_frag` after the HF calc to allow the scanner to still run:
```
if fd_info.kind in ("reference", "scanner_point"):
    # placeholder energy to allow Gradients.as_scanner() to return (energy, gradient)
    return mf.e_tot
```

tests/molbe_h6_scanner_test.py
bhanscam: See reply to previous comment. If you return the HF energy in the scanner, include that in the test as well as the gradient.

src/quemb/molbe/scanner.py
cariluo: I'm not sure `mc.e_tot + mf.energy_nuc()` includes the HF energy for the full system (not just the HF energy of the fragment). What I want is full system HF (and full system nuclear energy) plus fragment correlation.

src/quemb/molbe/scanner.py
```diff
@@ -150,6 +213,132 @@ def energy_be(mol, energy_args=None, fd_info=None):
...
+    redo_scratch = WorkDir(
+        mybe.scratch_dir / "redo_eri",
+        cleanup_at_end=True,
+        ensure_empty=True,
+    )
+    tmp_eri_file = redo_scratch / "eri.h5"
+
+    try:
+        fobj.TA = np.linalg.inv(mybe.S) @ S_cross @ ref_mybe.Fobjs[frag_idx].TA
+        fobj.dname = "redo" + str(frag_idx)
```
mscho527: I think we still have something like six ERI transformation calls sharing a same ERI file name. i.e., for fragment 0 (that contains atom 0), [+x, -x, +y, -y, +z, -z] displacements will all have "redo_0" dname and "eri.h5" eri file. We would want to give all of these separate names. On that note, maybe we should either copy the `fobj` or initialize it from scratch rather than modifying then reassigning its original values for cleanliness.

src/quemb/molbe/scanner.py
mscho527: Isn't your fragment HF energy identical to the full system HF energy? HF-in-HF embedding is exact by construction.

mscho527 (commented):
Two more minor requested changes, but I am okay with merging this PR once @bhanscam approves the changes based on the previous requests. I won't mark it as "Approved" yet so that @bhanscam has a chance to review, but will remove the blocking "change requested".

@cariluo what's the mypy error you see with graphgen? The latest GHA run doesn't seem to have mypy errors.

cariluo:
> @cariluo what's the mypy error you see with graphgen? The latest GHA run doesn't seem to have mypy errors.

I have removed the comment to suppress a type-checking error in graphfrag so that the error is apparent.

src/quemb/molbe/scanner.py
cariluo: If I do force embedding on a small polyethylene chain and print out the possible energy returns I get values like
```
mf.e_tot + mc.e_tot - fobj._mf.e_tot is -310.15434312373725
mc.e_tot + mf.energy_nuc() is 16.82353573729898
```

cariluo:
Summary of changes:

- no longer suppressing the type-checking error in `graphfrag.py`
- creating a unique dname and ERI file for each displacement
- deep copying the fragment object 
- replied to the conversation about what energy is returned in `energy_be_frag()`

mscho527:
Re: `graphfrag.py` mypy error, I suspect that something changed in the PyPlot colormap API recently, likely similar to what happened last in #208. You can create an issue (like #208) and suppress the mypy error again!

mscho527:
and tagging @ShaunWeatherly for heads up

bhanscam (commented):
Looks good! Happy to have this merge with main.

bhanscam (approved):

Issue #273, opened 2026-07-06
`mypy` error for `graphfrag.py`


cariluo:
There is a `mypy` error that shows up for `graphfrag.py` in `def export_graph()`:

```
Module has no attribute "get_cmap"  [attr-defined]
```

Suspected problem is that something changed in the PyPlot colormap API recently, likely similar to what happened last in #208.

cariluo:
Matplotlib's documentation specifically says that the old `matplotlib.cm.get_cmap()` is deprecated since 3.7 and recommends alternatives.

<img width="1202" height="1074" alt="Image" src="https://github.com/user-attachments/assets/64e9dd2b-3eb3-4ab3-9b3d-430ee682f6f4" />

PR #274, opened 2026-07-16
Fixed issue 258: updated dependency for chemcoord

hannahHYL:
Tested that changing to `pandas >= 3.0.0 chemcoord >= 2.2.0` still passes the tests. Preferred dependencies on pyproject.toml changed to `chemcoord>=2.2.0`. Per [chemfrag all-to-all connectivity test fails
 #258](https://github.com/troyvvgroup/quemb/issues/258) comment from @mcocdawc, pandas isn't on the dependency list. 

Resolves #258

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "numba",
     "ordered-set",
     "libdmet @ git+https://github.com/gkclab/libdmet_preview.git",
-    "chemcoord @ git+https://github.com/mcocdawc/chemcoord.git",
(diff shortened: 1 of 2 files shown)


mscho527:
The mypy typing error
```
src/quemb/molbe/graphfrag.py:201: error: Module has no attribute "get_cmap"  [attr-defined]
```
is known and is being tracked in #273 

If you change `graphfrag.py` like Carina did in #272, this will be suppressed. Could you add this to this PR?

mscho527 (approved):
Looks good, thanks!

PR #275, opened 2026-08-03
make fcidump work with pathlib:Path

### LINKED ISSUE #6: make fcidump work with `pathlib:Path`
the `molbe.misc.be2fcidump` function unnecessarily requires a `str` and does not work with `pathlib:Path` . Change that

cariluo:
This PR addresses issue #6


cariluo:
I see that the associated test is known to fail, but I hope that these changes address the specific concern in #6. Not urgent, but ready for review!

mscho527:
I think the fix should let the tests pass, right? Can you enable those tests?

cariluo:
`test_embedding()` in `molbe_io_fcidump_test.py` fails with value`1.1629791680047434` when on the `main` branch.

`test_embedding()` in `molbe_io_fcidump_test.py` fails with the same value, `1.1629791680047434` when on the branch `make_fcidump_work_with_Path`.

Issue #6 is a type-related issue, but I suspect there is another underlying issue that is causing `test_embedding()` to fail. 

I will do more debugging!

mscho527:
there is a chance that the target values in the tests are wrong (due to fixes we made in other parts of the code)
since the test is testing the io capacity, we just need to ensure that the io portion is correct
one thing to check, perhaps, is whether the target h1 and h2 matrix matches the 1- and 2-body matrix that you can get from the fragment fock and eri

cariluo:
I have added a test that I hope addresses the IO question, and it passes. Happy to make any revisions!

tests/molbe_io_fcidump_test.py
```diff
@@ -72,6 +130,16 @@ def test_fragment_mo() -> None:
     verify_fcidump_writing("fragment_mo")
```
mscho527: Let's remove these tests and the helper function, as these are obsolete and the new tests cover what we intend to cover

mscho527 (approved):
One minor modification request, but i think this is good otherwise!

tests/molbe_io_fcidump_test.py
mscho527: oh and the reference data we no longer use as well

